PR #5, opened 2021-12-16
First Pass Bulk Data

Jack-Fraser16:
NOTE: Most recent updates in separate comment below. 

I would say this PR represents about 90% completion with the First Pass. Some notes on what still needs to be completed:

TODO:
- [x] Group 3 -- There's a bug I'm chasing down that I believe comes from checking slices. Really did not get that deep into it before posting this, but its affecting about seven of the group 3 tests in the same way -- "No method exists for due[x]". Testing against the reference server with group_id 1 should reproduce the issue. 
- [x] Group 3 -- Provenance/Medication/Location tests
- [x] Unit Tests --  Not a 100% sure what to do about unit tests. I still need to write one or two more for BulkDataUtils -- however, the rest of the functionality was ported over from program, so I think it's solid (unless you think it's worth re-testing those). My big question though for writing unit tests for Group 3 is where I can find reference resources that fully conform to profiles. Is it worth bringing those in just for unit tests? If yes, should be pretty straightforward (if tedious) to write those unit tests. 

WAITING:
- [ ] TLS Tester Class needs to be implemented and TLS tests added to each group. Should be quick once the class is available.
- [ ] Group 2 and 3 tests use delete and stream functionality, which I have an [open PR](https://github.com/inferno-framework/inferno-core/pull/67) for in inferno-core. I wrote a very basic mock 'stream' request in BulkDataUtils, but this will ultimately be deleted once stream is available within the client. The singular delete test was just commented out for the time being.

Final notes: I pivoted away from passing previously made requests between individual tests because I thought it would be useful to be able to run each individual test with it checking for its required inputs -- I did not realize this would all end up being run as a large group anyways. This decision is mostly evident in Group 1 and a little bit in Group 2, so I'm aware the organization is a bit clunky. I'm more than happy to go back and re-organize, but before doing that wanted to get your thoughts -- is it worth the effort if it doesn't make a difference to how the group is being run?


lib/multi_patient_api/bulk_data_utils.rb
```diff
@@ -0,0 +1,383 @@
...
+								 else
+									 resolve_element_from_path(el.send(cur_path_part), path_ary.join('.'))
+								 end
+			return el_found if el_found.present? || el_found == false
+		end
+
+		nil
+	end
+
+	def find_slice_by_values(element, values)
```
Jammjammjamm, line 206: I am hoping that all of this must support checking can be handled by the `USCore` must support checking code.

lib/multi_patient_api/bulk_data_utils.rb
```diff
@@ -0,0 +1,383 @@
...
+			end
+		end 
+
+		line_count
+	end 
+
+	# Determine whether the file items in bulk_status_output contain resources 
+	#	that conform to the given profiles. 
+	# 
+	# @param klass [FHIR ResourceType]
```
Jammjammjamm, line 356: This is really a resource type rather than a class. Isn't it a string?

spec/multi_patient_api/bulk_data_authorization_spec.rb
```diff
@@ -0,0 +1,238 @@
...
+  # TODO: After TLS tester class is implemented, create this test
+  describe 'endpoint TLS tests' do
+    
+  end
+
+  describe '[Invalid grant_type] test' do
+    let(:runnable) { group.tests[1] }
+    let(:client_assertion) { create_client_assertion(client_assertion_input) }
+    let(:body) do
+      {
```
Jammjammjamm, line 53: I recommend having a set of good inputs, and changing whatever property is needed for this test so that it is clear what is general and what is test specific:
```
let(:test_specific_input) do
  general_input.merge(some_key: test_specific_value)
end
```

spec/multi_patient_api/bulk_data_authorization_spec.rb
```diff
@@ -0,0 +1,238 @@
+require_relative '../../lib/multi_patient_api/bulk_data_authorization.rb'
+require_relative '../../lib/multi_patient_api/bulk_data_utils.rb'
+
+RSpec.describe MultiPatientAPI::BulkDataAuthorization do
+  include BulkDataUtils
```
Jammjammjamm, line 5: It seems strange to include this module in the tests.

spec/multi_patient_api/bulk_data_authorization_spec.rb
```diff
@@ -0,0 +1,238 @@
...
+
+    it 'fails when token endpoint allows invalid grant_type' do
+      stub_request(:post, bulk_token_endpoint)
+        .with(body: body)
+        .to_return(status: 200, body: "", headers: {})
+
+      allow_any_instance_of(runnable).to receive(:create_client_assertion).and_return(client_assertion)
+      result = run(runnable, default_input)
+      
+      expect(result.result).to eq('fail')
```
Jammjammjamm, line 80: For negative tests, you should also make an assertion on the result message to make sure that the test is failing for the correct reason.

lib/multi_patient_api/bulk_data_utils.rb
```diff
@@ -0,0 +1,383 @@
+require 'pry'
+require 'json/jwt'
+module BulkDataUtils
+
+	include Inferno::DSL::Assertions
```
Jammjammjamm, line 5: This shouldn't be necessary since these methods will be used in tests which already include the assertions.

Gemfile
```diff
@@ -18,6 +18,7 @@ gem 'rake'
 gem 'rest-client'
 gem 'rubyzip'
 gem 'sqlite3'
+gem 'json-jwt'
```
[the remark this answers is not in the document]
Jack-Fraser16: A little uncertain where to place these within the gemspec file. Should they be added as runtime dependencies?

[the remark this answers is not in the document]
Jack-Fraser16: Also, there's a larger 'TODO' for all the gems - should I move them all over?

Jack-Fraser16:
For Group 3, I was able to pull in metadata directly from the US Core Tests -- except for the location and medication profiles, so I included copies of them under the multi-patient-api/metadata folder. Would love to know if there is a cleaner way to grab them from US Core. For Unit Tests for Group 3, I downloaded copies of the resources (in ndjson files) from the reference server and used them to test a mix of missing elements, missing slices, and missing resources. I also broke up the BulkDataUtils module into three separate modules, one for each group. For the ValidationUtils module, I'm unsure what to do with the `determine_profile` and `predefined_device_type?`. Neither of these methods is being used currently: `determine_profile` is basically `guess_profile` from inferno-program, which @arscan and I discussed being pretty useless for core. My thoughts in including `determine_profile` was that it went hand-in-hand with `predefined_device_type?` -- and I am also uncertain whether we need  `predefined_device_type?` in core. I definitely need some guidance on these, as well as the reasoning behind  counting the number of encounter resources (this also happens in program, and I did not include it as every resource is counted.)  The BulkDataUtils unit tests aren't really testing anything but the new stream - most other methods in that file are pulled verbatim from program or rely heavily on USCore MustSupport. 

One thing to note with testing against the reference server is that the Observation test fails - which I believe it fails correctly as there is a _http://hl7.org/fhir/StructureDefinition/bmi_ resource that doesn't conform to the profile. It's the 137th resource, so when this test fails it takes a WHILE for the next test to begin. It's something I noticed but haven't investigated, but my guess would be that it's taking a while for all the validation errors to come through and that's potentially causing the slow-down. Once stream is merged in, I'll play around with the `line_collection` variable and the stored request so that only the first 100 validated resources are being shown to the user.

Gemfile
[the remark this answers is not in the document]
Jammjammjamm: Yes, these are runtime dependencies. I wouldn't move the others over as part of this ticket.

Jack-Fraser16:
Added profile guessing and also re-organized the code a bit to make it more readable. Now, only missing [delete](https://github.com/inferno-framework/inferno-core/pull/67) and tls_tester (pr coming soon).

spec/multi_patient_api/resources/AllergyIntolerance.ndjson
```diff
@@ -0,0 +1,10 @@
+{"id":"4518502e-6062-790a-8583-3c9cf2ee288b","meta":{"profile":["http://hl7.org/fhir/us/core/StructureDefinition/us-core-allergyintolerance"]},"clinicalStatus":{"coding":[{"system":"http://terminology.hl7.org/CodeSystem/allergyintolerance-clinical","code":"inactive"}]},"verificationStatus":{"coding":[{"system":"http://terminology.hl7.org/CodeSystem/allergyintolerance-verification","code":"confirmed"}]},"type":"allergy","category":["food"],"criticality":"low","code":{"coding":[{"system":"http://snomed.info/sct","code":"300916003","display":"Latex allergy"}],"text":"Latex allergy"},"patient":{"reference":"Patient/e91975f5-9445-c11f-cabf-c3c6dae161f2"},"recordedDate":"1941-10-14T18:33:18-05:00","reaction":[{"manifestation":[{"coding":[{"system":"http://snomed.info/sct","code":"271807003","display":"skin rash"}],"text":"skin rash"}]}],"resourceType":"AllergyIntolerance"}
```
Jammjammjamm, line 1: These files should live in `spec/fixtures`.

spec/multi_patient_api/bulk_export_validation_tester_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    it 'applies process_header once, upon reception of a single, one-line chunk' do
+      stub_request(:get, url.to_s)
+        .to_return(status: 200, body: basic_body, headers: { 'Application' => 'expected' })
+
+      group.stream_ndjson(url, {}, generic_block, process_response_block)
+
+      expect(streamed_response).to eq(['expected'])
+    end
+
+    it 'applies process_header several times, upon reception of multiple chunks' do
```
Jammjammjamm, line 51: Should this test be implemented or removed?

spec/multi_patient_api/bulk_export_validation_tester_spec.rb
```diff
@@ -0,0 +1,72 @@
...
...l\",\"family\":\"van de Heuvel\",\"given\":[\"Pieter\"],\"suffix\":[\"MSc\"]}],\"telecom\":[{\"system\":\"phone\",\"value\":\"0648352638\",\"use\":\"mobile\"},{\"system\":\"email\",\"value\":\"p.heuvel@gmail.com\",\"use\":\"home\"}],\"gender\":\"male\",\"birthDate\":\"1944-11-17\",\"deceasedBoolean\":false,\"address\":[{\"use\":\"home\",\"line\":[\"Van Egmondkade 23\"],\"city\":\"Amsterdam\",\"postalCode\":\"1024 RJ\",\"country\":\"NLD\"}],\"maritalStatus\":{\"coding\":[{\"system\":\"http://terminology.hl7.org/CodeSystem/v3-MaritalStatus\",\"code\":\"M\",\"display\":\"Married\"}],\"text\":\"Getrouwd\"},\"multipleBirthBoolean\":true,\"contact\":[{\"relationship\":[{\"coding\":[{\"system\":\"http://terminology.hl7.org/CodeSystem/v2-0131\",\"code\":\"C\"}]}],\"name\":{\"use\":\"usual\",\"family\":\"Abels\",\"given\":[\"Sarah\"]},\"telecom\":[{\"system\":\"phone\",\"value\":\"0690383372\",\"use\":\"mobile\"}]}],\"communication\":[{\"language\":{\"coding\":[{\"system\":\"urn:ietf:bcp:47\",\"code\":\"nl\",\"display\":\"Dutch\"}],\"text\":\"Nederlands\"},\"preferred\":true}],\"managingOrganization\":{\"reference\":\"Organization/f001\",\"display\":\"Burgers University Medical Centre\"}}")
```
Jammjammjamm, line 20: This should live in a fixture file.

spec/multi_patient_api/bulk_data_group_export_validation_spec.rb
```diff
@@ -0,0 +1,514 @@
...
+    end
+
+    it 'skips when requires_access_token is not provided' do
+      result = run(runnable, { status_output: status_output })
+
+      expect(result.result).to eq('skip')
+      expect(result.result_message).to eq('Could not verify this functionality when requiresAccessToken is not provided')
+    end
+
+    # TODO: Figure out why this behaves weirdly. Suspicion: false is stored serially, and then interpreted as a truthy value.
```
Jammjammjamm, line 57: Does this need to be figured out? Seems like a problem.

spec/multi_patient_api/bulk_data_group_export_validation_spec.rb
```diff
@@ -0,0 +1,514 @@
...
+
+    it 'passes when the input patient ids do not match those stored' do
+      scratch[:patient_ids_seen] = ['one_id', 'two_id', 'three_id']
+      result = run(runnable, input)
+
+      expect(result.result).to eq('pass')
+    end
+  end
+
+  describe '[AllergyIntolerance resources returned conform to the US Core AllergyIntolerance Profile] test' do
```
Jammjammjamm, line 233: Why are AllergyIntolerance, CarePlan, DiagnosticReport, and Observation specifically tested? Should more tests be done on the tester class instead?

spec/multi_patient_api/bulk_export_validation_tester_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  extend Inferno::DSL::Configurable
+
+  include BulkExportValidationTester
+
+  def test_session_id
+    nil
+  end
+end
+
+RSpec.describe BulkExportValidationTesterClass do
```
Jammjammjamm, line 14: This should be describing `BulkExportValidationTester`.

lib/multi_patient_api/bulk_export_validation_tester.rb
```diff
@@ -0,0 +1,163 @@
+require_relative './profile_guesser'
+
+module BulkExportValidationTester
+  include USCore::MustSupportTest
+  include ProfileGuesser
+
+  MAX_NUM_COLLECTED_LINES = 100
+  MIN_RESOURCE_COUNT = 2
+
+  def observation_profiles
```
Jammjammjamm, line 10: These seem like metadata rather than profiles.

lib/multi_patient_api/bulk_export_validation_tester.rb
```diff
@@ -0,0 +1,163 @@
...
+        resource = FHIR.from_contents(line)
+      rescue StandardError
+        skip "Server response at line \"#{line_count}\" is not a processable FHIR resource."
+      end
+
+      skip_if resource.resourceType != resource_type,
+              "Resource type \"#{resource.resourceType}\" at line \"#{line_count}\" does not match type defined in output \"#{resource_type}\")"
+
+      determine_profile(resource).each { |profile| resources[profile] << resource }
+      patient_ids_seen << resource.id if resource_type == 'Patient'
```
Jammjammjamm, line 128: There's no accounting for duplicate patient ids.

lib/multi_patient_api/bulk_export_validation_tester.rb
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    stream_ndjson(url, build_headers(requires_access_token), process_line, process_headers)
+    validate_conformance(resources)
+
+    line_count
+  end
+
+  def perform_bulk_export_validation
+    skip_if !status_output, 'Could not verify this functionality when Bulk Status Output is not provided'
```
Jammjammjamm, line 148: I'd recommend using `.blank?` rather than `!` to check for falsiness.

Jammjammjamm (commented):
Rubocop needs to pass.

lib/multi_patient_api/profile_guesser.rb
```diff
@@ -0,0 +1,63 @@
...
+  def resource_contains_category(resource, category_code, category_system = nil)
+    resource&.category&.any? do |category|
+      category.coding&.any? do |coding|
+        coding.code == category_code && (category_system.blank? || coding.system.blank? || category_system == coding.system)
+      end
+    end
+  end
+
+  def guess_profile(resource)
+    resource.to_hash.dig('meta', 'profile') || begin
```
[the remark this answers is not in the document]
Jack-Fraser16: Okay.

lib/multi_patient_api/profile_guesser.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        return extract_profile('Bodytemp') if observation_contains_code(resource, '8310-5')
+
+        return extract_profile('Bodyweight') if observation_contains_code(resource, '29463-7')
+
+        return extract_profile('Heartrate') if observation_contains_code(resource, '8867-4')
+
+        return extract_profile('Resprate') if observation_contains_code(resource, '9279-1')
+
+        []
```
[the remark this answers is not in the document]
Jack-Fraser16: I was specifically thinking of the case when a resources `meta.profile` is an array of profiles -- possibly indicating that `must_supports` from different profiles are exemplified within that one resource. And if we're certifying based off returned resources demonstrating all `must_supports`, then only validating against one profile might mean we fail the following profiles.

E.g. a singular `DiagnosticReport` resource that conforms to both Note and Lab profiles and exemplifies a `must_support` from each - if we only validate against Note, then Lab may fail cause we'll never account for that Lab `must_support` also included.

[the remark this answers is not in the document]
Jammjammjamm: It may be technically possible in the spec, but I don't know that it actually makes real world sense to be both a lab report and a clinical note. That is not something that is currently supported in program edition.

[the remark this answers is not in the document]
Jack-Fraser16: Okay, I'll remove the arrays.

spec/multi_patient_api/bulk_export_validation_tester_spec.rb
Jack-Fraser16: Didn't actually use this resource in the tests :/

Jammjammjamm: The true power of code reviews!

spec/multi_patient_api/bulk_export_validation_tester_spec.rb
Jack-Fraser16: Removing it, unfortunately. The Faraday streaming capability is not great, and only gives access to the piecewise streamed-in chunks -- meaning I can't think of a way to check the response headers until after the full response is streamed in. So, can only check headers once after full response is received, which is less ideal than if we knew off the bat that say the content-type was wrong.

lib/multi_patient_api/bulk_export_validation_tester.rb
Jack-Fraser16: I'm with you but naming gets complicated because I need to leave `metadata` available for the calls to it within `perform_must_support_tests`. I'll try to get more descriptive.

spec/multi_patient_api/bulk_data_group_export_validation_spec.rb
Jack-Fraser16: Short answer: yes, I need to write more tests for the **tester** class. For the bulk_data_group_export_validation tests though, I think these are good. AllergyIntolerance and CarePlan just came up first and I think those tests cover most single-profile resource testing. I did DiagnosticReport and Observation tests specifically because those are the multiple profile cases, so I figured it'd be worth going deeper on those. Overall, it's my understanding that most single-profile resources like AllergyIntolerance and CarePlan will have the same behavior as what is tested  - we don't do any special profile guessing, so I'm not sure what would be significantly different. Feel free to disagree. Though, on second thought, I'm going to add some device tests as well.

spec/multi_patient_api/bulk_data_group_export_validation_spec.rb
Jack-Fraser16: Was passing in bool when it just wanted a string

Jack-Fraser16:
TODO: Clean-up Rubocop (this will be last thing I do before being merged) and confirm the cause of the Observation test failure when being run against the reference server.

Jack-Fraser16:
Cause of Observation test failure being dealt with by @Jammjammjamm.

Gemfile
```diff
@@ -20,6 +20,7 @@ gem 'rubocop', '~> 1.9'
 gem 'rubocop-rspec', require: false
 gem 'rubyzip'
 gem 'sqlite3'
+gem 'ndjson'
```
Jammjammjamm, line 23: This needs to go in the gemspec.

lib/g10_certification_test_kit/authorization_request_builder.rb
```diff
@@ -0,0 +1,85 @@
+require 'json/jwt'
+
+class AuthorizationRequestBuilder
+  def self.build(...)
+    new(...).authorization_request
+  end
+
+  def self.bulk_data_jwks
+    @bulk_data_jwks ||= JSON.parse(File.read(File.join(File.dirname(__FILE__), 'bulk_data_jwks.json')))
```
Jammjammjamm, line 9: You can use `__dir__` to get the directory of this file.

lib/g10_certification_test_kit/authorization_request_builder.rb
```diff
@@ -0,0 +1,85 @@
+require 'json/jwt'
+
+class AuthorizationRequestBuilder
```
Jammjammjamm, line 3: This needs to be in the `G10CertificationTestKit` module.

lib/g10_certification_test_kit/bulk_export_validation_tester.rb
```diff
@@ -0,0 +1,170 @@
...
+    end
+    ["USCore::#{resource_type}Group".constantize.metadata]
+  end
+
+  def metadata_list
+    @metadata_list ||= determine_metadata
+  end
+
+  def metadata
+    scratch[:metadata] ||= []
```
Jammjammjamm, line 39: No need for scratch. This could just be an `attr_reader` for a `@metadata` instance variable.

lib/g10_certification_test_kit/bulk_export_validation_tester.rb
```diff
@@ -0,0 +1,170 @@
+require_relative 'profile_guesser'
+
+module BulkExportValidationTester
```
Jammjammjamm, line 3: This needs to be in the `G10CertificationTestKit` module.

lib/g10_certification_test_kit/export_kick_off_performer.rb
```diff
@@ -0,0 +1,8 @@
+module ExportKickOffPerformer
```
Jammjammjamm, line 1: This needs to be in the `G10CertificationTestKit` module.

lib/g10_certification_test_kit/profile_guesser.rb
```diff
@@ -0,0 +1,66 @@
+module ProfileGuesser
```
Jammjammjamm, line 1: This needs to be in the `G10CertificationTestKit` module.

.gitignore
```diff
@@ -12,6 +12,7 @@
 .settings/org.eclipse.wst.jsdt.ui.superType.container
 .settings/org.eclipse.wst.jsdt.ui.superType.name
 .idea
+.package-lock.json
```
Jammjammjamm, line 15: Wrong filename.

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
Jammjammjamm: This file still needs to be removed.

lib/g10_certification_test_kit/bulk_data_group_export.rb
```diff
@@ -0,0 +1,259 @@
...
+      include ExportKickOffPerformer
+
+      run do
+        perform_export_kick_off_request
+        assert_response_status(202)
+
+        polling_url = response[:headers].find { |header| header.name == 'content-location' }.value
+        assert polling_url.present?, 'Export response header did not include "Content-Location"'
+
+        # delete(polling_url) TODO: Uncomment after core PR is merged in
```
Jammjammjamm, line 254: This was merged.

.rubocop.yml
```diff
@@ -31,6 +31,8 @@ Style/WordArray:
 Metrics/CyclomaticComplexity:
   Exclude:
     - 'lib/inferno/terminology/**'
+    - 'lib/g10_certification_test_kit/profile_guesser.rb'
+    - 'lib/g10_certification_test_kit/bulk_export_validation_tester.rb'
```
Jammjammjamm, line 35: You should disable this rule on specific lines rather than for entire files.

g10_certification_test_kit.gemspec
```diff
@@ -7,9 +7,11 @@ Gem::Specification.new do |spec|
   spec.description   = 'G10 Certification Tests for Inferno'
   spec.homepage      = 'https://github.com/inferno_framework/g10-certification-test-kit'
   spec.license       = 'Apache-2.0'
-  spec.add_runtime_dependency 'inferno_core', '~> 0.1.2'
   spec.add_runtime_dependency 'bloomer', '~> 1.0.0'
   spec.add_runtime_dependency 'colorize', '~> 0.8.1'
+  spec.add_runtime_dependency 'inferno_core', '~> 0.1.2'
+  spec.add_runtime_dependency 'json-jwt'
```
Jammjammjamm, line 13: It's a good idea to specify versions on these so we don't get bitten by breaking changes. You can generally do `~> whatever_version_is_in_Gemfile.lock`.

lib/g10_certification_test_kit/bulk_data_authorization.rb
```diff
@@ -0,0 +1,217 @@
...
+    input :bulk_scope,
+          title: 'Bulk Data Scopes',
+          description: 'Bulk Data Scopes provided at registration to the Inferno application.',
+          default: 'system/*.read'
+    input :bulk_encryption_method,
+          title: 'Encryption Method',
+          description: 'The server is required to suport either ES384 or RS384 encryption methods
+                        for JWT signature verification. Select which method to use.',
+          default: 'ES384'
+    # TODO: Make radio button
```
Jammjammjamm, line 38: Make it a radio button.

lib/g10_certification_test_kit/bulk_export_validation_tester.rb
```diff
@@ -0,0 +1,172 @@
...
+
+    def metadata_list
+      @metadata_list ||= determine_metadata
+    end
+
+    def patient_ids_seen
+      scratch[:patient_ids_seen] ||= []
+    end
+
+    # TODO: Delete this once core functionality is merged in
```
Jammjammjamm, line 45: This was merged.

spec/g10_certification_test_kit/bulk_data_group_export_spec.rb
```diff
@@ -0,0 +1,321 @@
...
+    end
+
+    it 'skips when polling_url is not provided' do
+      result = run(runnable)
+
+      expect(result.result).to eq('skip')
+      expect(result.result_message).to eq('Server response did not have Content-Location in header')
+    end
+
+    # NOTE: Test commented out as it takes three minutes to fail
```
Jammjammjamm, line 195: Check out TimeCop, it's how you deal with these situations: https://github.com/travisjeffery/timecop

lib/g10_certification_test_kit/bulk_data_group_export.rb
```diff
@@ -248,10 +248,12 @@ class BulkDataGroupExport < Inferno::TestGroup
         perform_export_kick_off_request
         assert_response_status(202)
 
-        polling_url = response[:headers].find { |header| header.name == 'content-location' }.value
+        polling_url = response[:headers].find { |header| header.name == 'content-location' }&.value
```
Jammjammjamm, line 251: You should instead use `request.response_header('content-location')&.value`, which handles case-insensitive matching of headers.

spec/g10_certification_test_kit/bulk_data_group_export_spec.rb
```diff
@@ -192,17 +193,17 @@ def run(runnable, inputs = {})
...
-    #   stub_request(:get, "#{polling_url}")
-    #     .with(headers: { 'Authorization' => "Bearer #{bearer_token}" } )
-    #     .to_return(status: 202)
+    it 'skips when server only returns "202 Accepted", and not "200 OK" in the allowed timeframe' do
+      stub_request(:get, polling_url.to_s)
+        .with(headers: { 'Authorization' => "Bearer #{bearer_token}" })
+        .to_return(status: 202)
 
-    #   result = run(runnable, base_input)
+      allow_any_instance_of(runnable).to receive(:sleep) { |_o, time| Timecop.travel(Time.now + time) }
```
Jammjammjamm, line 201: If you're just mocking `sleep`, you don't need timecop. You can just make `sleep` do nothing.

lib/g10_certification_test_kit/bulk_data_authorization.rb
```diff
@@ -0,0 +1,229 @@
...
+
+        The OAuth 2.0 Authorization Framework describes the proper response for an
+        invalid request in the client credentials grant flow:
+
+        ```
+        If the request failed client authentication or is invalid, the authorization server returns an
+        error response as described in [Section 5.2](https://tools.ietf.org/html/rfc6749#section-5.2).
+        ```
+      DESCRIPTION
+      # link 'http://hl7.org/fhir/uv/bulkdata/authorization/index.html#protocol-details'
```
radamson, line 83: Should this link be included directly in the description?

lib/g10_certification_test_kit/bulk_data_group_export.rb
```diff
@@ -0,0 +1,258 @@
...
+        capability_statement.rest&.each do |rest|
+          groups = rest.resource&.select { |resource| resource.type == 'Group' }
+
+          pass if groups&.any? do |group|
+            group.operation&.any? do |operation|
+              if operation.definition.is_a? String
+                operation.definition == definition
+              else
+                operation.definition&.flatten&.include?(definition)
+              end
```
radamson, line 68: `definition` is constrained `1..1` Cardinality-wise, so I don't think these checks are needed?

lib/g10_certification_test_kit/bulk_data_group_export.rb
```diff
@@ -0,0 +1,258 @@
...
+      DESCRIPTION
+      # link 'http://hl7.org/fhir/uv/bulkdata/export/index.html#bulk-data-kick-off-request'
+
+      include ExportKickOffPerformer
+
+      run do
+        skip_if bearer_token.blank?, 'Could not verify this functionality when bearer token is not set'
+
+        perform_export_kick_off_request(use_token: false)
+        assert_response_status(401)
```
radamson, line 94: The descriptions says `recommend using HTTP status code 401` but the test only checks for `401`

It looks like Inferno-Program allowed 400 or 401

https://github.com/onc-healthit/inferno-program/blob/a773c8ed407875b9614e11f692e51f350a6c2f6b/lib/modules/onc_program/bulk_data_export_sequence.rb#L188

https://github.com/onc-healthit/inferno-program/blob/a773c8ed407875b9614e11f692e51f350a6c2f6b/lib/app/utils/assertions.rb#L41

radamson (commented):
Overall looks really good! Some initial comments/questions:

lib/g10_certification_test_kit/bulk_data_authorization.rb
```diff
@@ -0,0 +1,229 @@
+require_relative 'authorization_request_builder'
+
+module G10CertificationTestKit
+  class BulkDataAuthorization < Inferno::TestGroup
+    title 'Bulk Data Authorization'
+    description <<~DESCRIPTION
+      Bulk Data servers are required to authorize clients using the
+      [Backend Service Authorization](http://hl7.org/fhir/uv/bulkdata/STU1/authorization/)
+      specification as defined in the [FHIR Bulk Data Authorization Guide](http://hl7.org/fhir/uv/bulkdata/STU1/).
```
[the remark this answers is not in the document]
Jammjammjamm: This is fine, but we do have a separate ticket to update all of the smart/bulk data links.

lib/g10_certification_test_kit/authorization_request_builder.rb
```diff
@@ -0,0 +1,87 @@
+require 'json/jwt'
+
+module G10CertificationTestKit
+  class AuthorizationRequestBuilder
+    def self.build(...)
+      new(...).authorization_request
+    end
```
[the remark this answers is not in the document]
Jammjammjamm: I don't have strong feelings either way.

lib/g10_certification_test_kit/bulk_data_group_export.rb
```diff
@@ -0,0 +1,258 @@
...
+      run do
+        assert status_response.present?, 'Bulk Data Server status response not found'
+
+        requires_access_token = JSON.parse(status_response)['requiresAccessToken'].to_s.downcase
+        output requires_access_token: requires_access_token
+
+        assert requires_access_token.present? && requires_access_token == 'true',
+               'Bulk Data file server access SHALL require access token'
+      end
+    end
```
radamson, line 232:
```suggestion
```

lib/g10_certification_test_kit/authorization_request_builder.rb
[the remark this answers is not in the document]
Jack-Fraser16: If it's all the same, I'm going to keep as is, just for the sake of getting this in.

lib/g10_certification_test_kit/bulk_data_group_export.rb
Jack-Fraser16: Good catch.

lib/g10_certification_test_kit/bulk_data_group_export.rb
```diff
@@ -0,0 +1,258 @@
...
+          seconds_used = Time.now - start + wait_time
+
+          break if response[:status] != 202 || seconds_used > timeout
+
+          sleep wait_time
+        end
+
+        skip "Server took more than #{timeout} seconds to process the request." if response[:status] == 202
+
+        assert response[:status] == 200, "Bad response code: expected 200, 202, but found #{response[:status]}."
```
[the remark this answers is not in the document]
Jack-Fraser16: With the exception of like two words, the message this assert was outputting was almost identical to `assert_response_status`. So, just going to use that.

lib/g10_certification_test_kit/bulk_data_authorization.rb
Jack-Fraser16: Not a 100% sure, but will defer to anyone with a strong opinion. I think the description gives a solid narrative view of why/what's being tested and though the link might add some specificity, I don't think it's crucial to point users in that direction -- especially if we're already linking an overview of the authorization documentation at the beginning of the group. Again though, I'm okay with whatever.

lib/g10_certification_test_kit/bulk_data_authorization.rb
```diff
@@ -0,0 +1,229 @@
...
+          title: 'Bulk Data Client ID',
+          description: 'Client ID provided at registration to the Inferno application.'
+    input :bulk_scope,
+          title: 'Bulk Data Scopes',
+          description: 'Bulk Data Scopes provided at registration to the Inferno application.',
+          default: 'system/*.read'
+    input :bulk_encryption_method,
+          title: 'Encryption Method',
+          description: 'The server is required to suport either ES384 or RS384 encryption methods
+                        for JWT signature verification. Select which method to use.',
```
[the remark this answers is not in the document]
Jack-Fraser16: Sounds good.

lib/g10_certification_test_kit/bulk_data_group_export.rb
Jack-Fraser16: Deleted this test.

lib/g10_certification_test_kit/bulk_data_authorization.rb
radamson: Inferno-Program had a `link` method for providing a reference to the spec. @Jammjammjamm is `link` getting carried over to core? If not it seems like this can either be deleted or incorporated into the description.

lib/g10_certification_test_kit/bulk_data_group_export.rb
Jack-Fraser16: Actually deleted this too hastily. It ended up breaking quite a few things, and I thought there would be a quick and easy fix, but now I think our best bet is to just get this PR in with this test and then figure out removing requiresAccessToken after.

Jack-Fraser16:
Added back in the requires_access_token test. Confident that I can get make updates with this test removed before Wednesday; however, an initial, lazy attempt to the remove the test broke a couple more things than I anticipated. Just for peace of my mind, I think it's best to get the PR merged in with the test (as it stands currently) and then update it in the next day or two.

lib/g10_certification_test_kit/bulk_data_authorization.rb
```diff
@@ -0,0 +1,229 @@
...
+    id :bulk_data_authorization
+
+    input :bulk_token_endpoint,
+          title: 'Backend Services Token Endpoint',
+          description: 'The OAuth 2.0 Token Endpoint used by the Backend Services specification
+                        to provide bearer tokens.',
+          default: 'https://inferno.healthit.gov/bulk-data-server/auth/token'
+    input :bulk_client_id,
+          title: 'Bulk Data Client ID',
+          description: 'Client ID provided at registration to the Inferno application.'
```
radamson, line 28:
```suggestion
          description: 'Client ID provided at registration to the Inferno application.'
          default: 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6InJlZ2lzdHJhdGlvbi10b2tlbiJ9.eyJqd2tzX3VybCI6Imh0dHBzOi8vaW5mZXJuby5oZWFsdGhpdC5nb3YvaW5mZXJuby8ud2VsbC1rbm93bi9qd2tzLmpzb24iLCJhY2Nlc3NUb2tlbnNFeHBpcmVJbiI6MTUsImlhdCI6MTU5OTE1NzgyMX0.-wulnE05BlY_Zcm5iP77Meqxr6iNiYxBsOADB5CGE8I'
```
Can we bring over the default from Program

lib/g10_certification_test_kit/bulk_data_group_export.rb
```diff
@@ -0,0 +1,252 @@
...
+    description <<~DESCRIPTION
+      Verify that system level export on the Bulk Data server follow the Bulk Data Access Implementation Guide
+    DESCRIPTION
+
+    id :bulk_data_group_export
+
+    input :bearer_token
+    input :bulk_server_url,
+          title: 'Bulk Data FHIR URL',
+          description: 'The URL of the Bulk FHIR server.'
```
radamson, line 15:
```suggestion
          description: 'The URL of the Bulk FHIR server.'
          default: 'https://inferno.healthit.gov/bulk-data-server/eyJlcnIiOiIiLCJwYWdlIjoxMDAwMCwiZHVyIjoxMCwidGx0IjoxNSwibSI6MSwic3R1Ijo0fQ/fhir'
```

radamson (approved):
Tested it out without issue. It does look like some of the defaults previously in Program were missing so I added some suggestions to add them back in

Screen Shot 2022-02-07 at 10 54 42 AM


Screen Shot 2022-02-07 at 10 55 40 AM

Not sure if we havea ticket for this already, but it would be great if the UI could indicate whch resources or http requests the warnings were related to (absolutely should be a separate task though)

PR #8, opened 2022-01-19
FI-1363: Limited access tests

Jammjammjamm:
This branch adds tests for unrestricted access and for the limited app launch.

### DIFF
--- a/lib/g10_certification_test_kit/resource_access_test.rb
+++ b/lib/g10_certification_test_kit/resource_access_test.rb
@@ -26,7 +26,7 @@ module G10CertificationTestKit
     end
 
     def search_param_value(name)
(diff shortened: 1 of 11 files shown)


lib/g10_certification_test_kit/resource_access_test.rb
```diff
@@ -0,0 +1,96 @@
+module G10CertificationTestKit
+  class ResourceAccessTest < Inferno::Test
+    id :g10_resource_access_test
+    input :patient_id, :received_scopes
+
+    title 'Access to resources are restricted properly based on patient-selected scope'
+    description %(
+      This test ensures that access to the a resource is granted or denied
```
Jack-Fraser16, line 8:
```suggestion
      This test ensures that access to the resource is granted or denied
```

PR #16, opened 2022-02-08
remove requiresAccessToken test and update unit tests

Jack-Fraser16:


lib/g10_certification_test_kit/bulk_data_group_export.rb
```diff
@@ -166,6 +166,7 @@ class BulkDataGroupExport < Inferno::TestGroup
           assert response_body.key?(key), "Complete Status response did not contain \"#{key}\" as required"
         end
 
+        output requires_access_token: response_body['requiresAccessToken'].to_s.downcase
```
radamson, line 169: Do we really have to turn this boolean into a string?  Since the output is stored as JSON it seems like this should be possible

https://github.com/inferno-framework/inferno-core/blob/6614f231150529b79d79003b2753018e9e7d7a24/lib/inferno/test_runner.rb#L96

@Jammjammjamm?

radamson (approved):
Code looks good and was able to test without issue:

Screen Shot 2022-02-08 at 12 16 55 PM

lib/g10_certification_test_kit/bulk_data_group_export.rb
radamson: I was trying to say you don't need any of `.to_s.downcase` since its a boolean in the JSON response, not a string, so letter case shouldn't matter.

```suggestion
        output requires_access_token: response_body['requiresAccessToken']
```

Jammjammjamm (changes requested):
This test needs to be updated to omit rather than skip if `requiresAccessToken` is false: https://github.com/inferno-framework/g10-certification-test-kit/blob/main/lib/g10_certification_test_kit/bulk_data_group_export_validation.rb#L44

lib/g10_certification_test_kit/bulk_data_group_export.rb
Jammjammjamm: Also, all inputs/outputs are stored as strings in the database.

radamson: Gotcha. I was misunderstanding the input/output saving and loading

Jammjammjamm (approved):

PR #20, opened 2022-02-17
FI-1450: Check configuration

Jammjammjamm:
This branch checks the available terminology resources and validator version. It needs a new core release to function, and I don't think it should be merged until the configuration UI is updated.

You can test it out by loading the g10 test kit in core.

**TODO:** Once a new core version is released, update the gemspec to require that new version.


arscan (approved):
I think we should go ahead and merge these in, as long as we pull in this https://github.com/inferno-framework/inferno-core/pull/113 so that our deployment is unaffected (because that has no errors).

Its quite a bit easier to work on the UI when we have real messages coming through

PR #21, opened 2022-02-18
FI-1461: Rename gem and update titles

Jammjammjamm:


lib/g10_certification_test_kit.rb
```diff
@@ -34,8 +34,8 @@ def required_inputs(prior_outputs = [])
 
 module G10CertificationTestKit
```
arscan, line 35: Probably kind of a pain, but should we change module names, file names, etc to be 100% consistent.

And I know we don't really have plans to publish this as a gem, but it seems like a good idea to have the `.rb` file match gem name, etc, right?

Jammjammjamm: You are right, I forgot to do that. I do plan to actually publish this as a gem.

arscan (approved):

PR #30, opened 2022-03-04
Omit location medication

Jack-Fraser16:
Allow Medication and Location tests in Bulk Data Export Validation to omit instead of skip if the respective resources are not returned.

### DIFF
--- a/lib/onc_certification_g10_test_kit/bulk_data_group_export_validation.rb
+++ b/lib/onc_certification_g10_test_kit/bulk_data_group_export_validation.rb
@@ -456,7 +456,7 @@ module ONCCertificationG10TestKit
     test do
       title 'Medication resources returned conform to the US Core Medication Profile'
       description <<~DESCRIPTION
-        This test verifies that the resources returned from bulk data export conform to the US Core profiles. This includes checking for missing data elements and value set verification. This test is omitted if bulk data export does not return any Medication resources.#{' '}
+        This test verifies that the resources returned from bulk data export conform to the US Core profiles. This includes checking for missing data elements and value set verification. This test is omitted if bulk data export does not return any Medication resources.
       DESCRIPTION
       # link 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-medication'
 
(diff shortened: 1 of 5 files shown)


lib/onc_certification_g10_test_kit/bulk_data_group_export_validation.rb
```diff
@@ -456,7 +456,7 @@ def resource_type
     test do
       title 'Medication resources returned conform to the US Core Medication Profile'
       description <<~DESCRIPTION
-        This test verifies that the resources returned from bulk data export conform to the US Core profiles. This includes checking for missing data elements and value set verification.
+        This test verifies that the resources returned from bulk data export conform to the US Core profiles. This includes checking for missing data elements and value set verification. This test is omitted if bulk data export does not return any Medication resources.#{' '}
```
arscan, line 459: Whats up with that `#{' '}`?  I've seen it in a few places of these descriptions.

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
```diff
@@ -152,13 +155,19 @@ def check_file_request(url) # rubocop:disable Metrics/CyclomaticComplexity
       line_count
     end
 
+    def omit_or_skip(message)
+      omit_if (OMIT_KLASS.include? resource_type), message
+      skip message
```
arscan, line 160: Why is this a skip?  If you they are sending the wrong type of resource in this file, I believe that this is a failure.  We use skip to mean 'haven't sent enough data for us to decide if you should pass', which doesn't apply here.

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
```diff
@@ -158,7 +163,11 @@ def perform_bulk_export_validation
               'Could not verify this functionality when Bearer Token is required and not provided'
 
       file_list = JSON.parse(status_output).select { |file| file['type'] == resource_type }
-      skip_if file_list.empty?, "No #{resource_type} resource file item returned by server."
+      message = "No #{resource_type} resource file item returned by server."
+      if file_list.empty?
+        omit_if (OMIT_KLASS.include? resource_type), message
+        skip message
+      end
```
arscan, line 170:
```suggestion
      if file_list.empty?
        message = "No #{resource_type} resource file item returned by server."
        omit_if (OMIT_KLASS.include? resource_type), message
        skip message
      end
```

arscan: Is message used outside of this if block? If not, probably should move it inside.

arscan (approved):
This looks fine at the code level, but I haven't had a chance to run it.  I think if it passes our reference server it is good to merge.

PR #31, opened 2022-03-04
FI-1462 matrix

yunwwang:
This PR create ONC Certification G10 Test Kit matrix for test procedures.


lib/onc_certification_g10_test_kit/tasks/generate_matrix.rb
```diff
@@ -0,0 +1,243 @@
...
+require 'inferno'
+require_relative '../../onc_certification_g10_test_kit'
+require_relative 'test_procedure'
+
+module ONCCertificationG10TestKit
+  module Tasks
+    class GenerateMatrix
+      include ONCCertificationG10TestKit
+
+      FILE_NAME = 'onc_certification_g10_matrix.xlsx'.freeze
```
arscan, line 12: It would be helpful to put version in the filename as well.

arscan (approved):
This looks good. I did a quick sanity check on a few of the parts and the content looks right to me.  A few things:
* Do we want to commit the xlsx in this repo?  We did in inferno-program, but we put it in a static folder and served that up so you could get to it while running the app.  But we could just rely on github releases to host this for us going forward.
* After https://github.com/onc-healthit/onc-certification-g10-test-kit/pull/29 is in we'll need to update the procedure.yml file to map in that test, and regenerate.
* You'll have to rerun this to regenerate the file after this one because some of the title are changed slightly: https://github.com/onc-healthit/onc-certification-g10-test-kit/pull/30]

Note: if @Jammjammjamm would like to put a pause on this, we don't technically need to merge it.  As long as we can get this xlsx file into the release as an attachment in GitHub early next week thats fine I think, as as long as we use this to regenerate the xlsx based on the final content of v2 that's good enough.

lib/onc_certification_g10_test_kit/tasks/generate_matrix.rb
arscan: n/m we are leaving it the same name.

PR #49, opened 2022-03-18
FI-1410: Update input forms

Jammjammjamm:
This branch cleans up the order and labeling of the inputs. This branch has to be tested by loading it in the inferno core repo, because it relies on functionality that hasn't yet been released.


arscan (approved):
Tried it out through inferno-core and it worked fine.

For PKCE -- should there technically be a prompt for each different launch?  I wonder if some systems might implement it for the public clients but not confidential ones.

Extending the limited access grant input to 'Expected Resource Grant for Limited Access Launch' would be helpful I think.

Issue #51, opened 2022-03-21
5.2.02 - Group Export Capability Statement Test Does Not Allow Custom OperationDefinitions


MadhurRaj:
Currently, the test for the capability statement expects the url to match the url of the Operation Definition from the IG.
[See Test](https://github.com/onc-healthit/onc-certification-g10-test-kit/blob/v2.0.0/lib/onc_certification_g10_test_kit/bulk_data_group_export.rb#L75)

This doesn't allow the possibility of using a custom OperationDefinition that is based on the IG, which is allowed by the FHIR Spec. The test is currently checking whether the server implements the exact Operation. 

This issue was discussed on the [Zulip Chat](https://chat.fhir.org/#narrow/stream/179309-inferno/topic/BDE-02.3A.20Bulk.20Data.20Server.20declares.20support.20for.20Group.20export).

yunwwang:
@MadhurRaj Does @chris-at-work 's changes match your operation definition?

PR #52, opened 2022-03-22
Relax regex on bulk data server group export url

chris-at-work:
Reported issue about having a server on a url that does not match our reference server and hardcoded fixtures.  I'm not sure how much of this URL is the _"base"_.

This is the regex: `%r{^http(s)?://.*/OperationDefinition/group-export(\|\S+)?$}`

Added a test, watched it pass/flip both ways on http|https, version number at the end, before this PR, the whole red/green loop.  So I'm confident in this fix but not on what we want to consider the base url.

Related: #51

### DIFF
--- a/lib/onc_certification_g10_test_kit/bulk_data_group_export.rb
+++ b/lib/onc_certification_g10_test_kit/bulk_data_group_export.rb
@@ -73,7 +73,7 @@ module ONCCertificationG10TestKit
             resource.type == 'Group' &&
               resource.respond_to?(:operation) &&
               resource.operation&.find do |operation|
-                name_match = operation.name == 'export'
+                name_match = ['export', 'group-export'].include?(operation.name)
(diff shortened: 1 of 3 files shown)


Jammjammjamm (approved):

lib/onc_certification_g10_test_kit/bulk_data_group_export.rb
```diff
@@ -73,7 +73,12 @@ class BulkDataGroupExport < Inferno::TestGroup
             resource.type == 'Group' &&
               resource.respond_to?(:operation) &&
               resource.operation&.find do |operation|
-                operation.definition&.match(%r{^http://hl7.org/fhir/uv/bulkdata/OperationDefinition/group-export(\|\S+)?})
+                name_match = ['export', 'group-export'].include?(operation.name)
```
arscan, line 76: I don't understand why we are allowing `group-export` in the name.  If we are, it seems to me that there isn't great consensus on what the right behavior should be, and we should just stop trying to enforce anything here...

Issue #60, opened 2022-03-29
Support redirect response for Bulk Data Output File Request 5.3.03 (FI-1522)


sstanciu5:
The current implementation in the ONC Certification test kit seems to read directly from the body of the Bulk Data Output File Request https://github.com/onc-healthit/inferno-program/blob/main/lib/modules/onc_program/bulk_data_group_validation_sequence.rb#L108.

This means that FHIR implementations that keep the export results on external hosting services and make use of HTTP redirects to   point the client to the external hosting location will fail step [5.3.03](https://inferno.healthit.gov/suites/test_sessions/04069015-570c-4ac3-a4bf-ae331c567adf#g10_certification-multi_patient_api) (and subsequent resource validation tests) of the Multi-Patient API test suite.

https://inferno.healthit.gov/suites/test_sessions/04069015-570c-4ac3-a4bf-ae331c567adf#g10_certification-multi_patient_api

<img width="1086" alt="image" src="https://user-images.githubusercontent.com/82027461/160651134-64ec3d9d-9c79-48b0-81bd-853417292ef3.png">

The Auth header should be removed from the request following the redirect (pre-signed URL).

Cerner has implemented redirect workflow that requires an access token to initiate the download request and responds with a pre-signed redirect URL. We have met with the ONC and they are in agreement that our implementation is in alignment with regulatory expectations.

Can the ONC test kit be updated to take into account a redirect status code & Location header and follow the header to download the results file ?

arscan:
This was also discussed in the SMART/FHIR community, and [this message](https://chat.fhir.org/#narrow/stream/179250-bulk-data/topic/301.20redirect/near/272431571) summarizes the outcome of discussions.  There isn't any guidance on behavior for authorized clients on redirects, but there probably should be.

Following redirects seems like a reasonable expected behavior of an HTTP client, so we can do that.  Regarding not sending the Authorization header, since there isn't any guidance about whether or not the client should be sending it in this context (clients are allowed to do either way according to HTTP/OAUTH/SMART), we could provide a toggle in the tester interface to determine how the client acts (send or do not send Authorization header if redirected).  This means that every bulk data client that expects to be flexible enough to be used against any g10-certified API will also need to have this configuration option.

Out of curiosity, are the initial files (that return redirects) served from the Binary resource, or are they outside the path of the fhir server that performed $export?

sstanciu5:
[...]

PR #72, opened 2022-04-06
FI-1522 Accept Redirect 301/302/303/307 for Bulk Data file download

yunwwang:
This PR fixes GitHub Issue #60 

The changes allows Bulk Data File download testing accept HTTP Redirect status 301/302/303/307

When requesting redirect endpoint, Bulk Data client does NOT send authorization header.

### DIFF
--- a/lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
+++ b/lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
@@ -58,7 +58,7 @@ module ONCCertificationG10TestKit
       headers
     end
 
-    def stream_ndjson(endpoint, headers, process_chunk_line, process_response)
+    def stream_ndjson(endpoint, headers, process_chunk_line, process_response) # rubocop:disable Metrics/CyclomaticComplexity
       hanging_chunk = String.new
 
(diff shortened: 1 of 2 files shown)


lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
```diff
@@ -74,6 +74,21 @@ def stream_ndjson(endpoint, headers, process_chunk_line, process_response)
 
       stream(process_body, endpoint, headers: headers)
 
+      max_redirect = 5
+
+      redirect_url = (response[:headers].find { |header| header.name.downcase == 'location' })&.value
```
Jammjammjamm, line 79:
```suggestion
      redirect_url = request.response_header('location')&.value
```

Jammjammjamm: This needs to happen in the `while` loop somewhere to handle multiple redirects, right?

Jammjammjamm (commented):
I think the cases we need to unit test are:
- single absolute redirect (done)
- single relative redirect
- multiple redirect

Jammjammjamm (approved):

PR #75, opened 2022-04-07
Relax 40X error specificity in Additional Tests

kayadanae:
Inferno's Tab 6: Additional Tests currently features two tests that evaluate very similar functionality.

**6.2.03 Token refresh fails after token revocation** ensures that, when given a refresh token that has been revoked and is no longer valid, the Health IT module does not issue access tokens.

**6.5.03 OAuth token exchange fails when supplied invalid code** ensures that, when given an authorization code that is invalid, the Health IT module does not issue access tokens.

Auth0, one of the most popular authorization and authentication providers, considers both of these `invalid_grant` errors, which yield 403 responses per their [documentation](https://auth0.com/docs/api/authentication#standard-error-responses) on Standard Error Responses.

It's in the best interest of both Inferno and the applications it evaluates to accept industry-standard HTTP responses so as not to force Health IT modules to insert home-rolled shims that intersect and massage auth responses, which have the potential to introduce unnecessary lag, complexity, and in the worst case vulnerabilities.

(In addition to adding 403 to both tests, I chose to keep 401 in the 6.2.03 allow list so as not to introduce breaking changes for modules who may be returning that error code already, and subsequently added 401 to 6.5.03 so that the test criteria would have parity.)

Thanks in advance for your consideration.

### DIFF
--- a/lib/onc_certification_g10_test_kit/smart_invalid_token_group.rb
+++ b/lib/onc_certification_g10_test_kit/smart_invalid_token_group.rb
@@ -158,7 +158,7 @@ module ONCCertificationG10TestKit
 
         post(smart_token_url, body: oauth2_params, name: :token, headers: oauth2_headers)
 
-        assert_response_status(400)
+        assert_response_status([400, 401, 403])
       end
     end
 
--- a/lib/onc_certification_g10_test_kit/token_revocation_group.rb
+++ b/lib/onc_certification_g10_test_kit/token_revocation_group.rb
@@ -126,7 +126,7 @@ module ONCCertificationG10TestKit
 
         post(smart_token_url, body: oauth2_params, headers: oauth2_headers)
 
-        assert_response_status([400, 401])
+        assert_response_status([400, 401, 403])
       end
     end
   end


arscan:
Thanks for letting us know about the discrepancy between our interpretation of the specs and Auth0's implementation.  And also thanks for doing the legwork of identifying what code would need to change.

I agree that a very bad outcome here would be if developers had to write/maintain code & services to work around this difference to make Auth0-based implementations pass these tests.

The language where we got this test is from the [OAuth 2.0 RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2):
```
5.2  Error Response

   The authorization server responds with an HTTP 400 (Bad Request)
   status code (unless specified otherwise) and includes the following
   parameters with the response:

   error
         REQUIRED.  A single ASCII error code from the
         following:

         invalid_request
               The request is missing a required parameter, includes an
               unsupported parameter value (other than grant type),
               repeats a parameter, includes multiple credentials,
               utilizes more than one mechanism for authenticating the
               client, or is otherwise malformed.

         invalid_client
               Client authentication failed (e.g., unknown client, no
               client authentication included, or unsupported
               authentication method).  The authorization server MAY
               return an HTTP 401 (Unauthorized) status code to indicate
               which HTTP authentication schemes are supported.  If the
               client attempted to authenticate via the "Authorization"
               request header field, the authorization server MUST
               respond with an HTTP 401 (Unauthorized) status code and
               include the "WWW-Authenticate" response header field
               matching the authentication scheme used by the client.

         invalid_grant
               The provided authorization grant (e.g., authorization
               code, resource owner credentials) or refresh token is
               invalid, expired, revoked, does not match the redirection
               URI used in the authorization request, or was issued to
               another client.

         unauthorized_client
               The authenticated client is not authorized to use this
               authorization grant type.

         unsupported_grant_type
               The authorization grant type is not supported by the
               authorization server.
```

So we expect it to be a 400 unless it is an invalid client.

Perhaps having this be a strict requirement wasn't the intent of the spec writers though (no must/shall on the 400), and maybe there is no reason to not allow other codes.  Are we interpreting 'unless otherwise specified' incorrectly maybe?

Are there other major implementations out there that do something other than 400?  I'm not sure that a single implementation defines 'industry standard', though I suspect Auth0 carries a lot of weight here.

arscan:
Thanks for letting us know about the discrepancy between our interpretation of the specs and Auth0's implementation.  And also thanks for doing the legwork of identifying what code would need to change.

I agree that a very bad outcome here would be if developers had to write/maintain code & services to work around this difference to make Auth0-based implementations pass these tests.

The language where we got this test is from the [OAuth 2.0 RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2):
```
5.2  Error Response

   The authorization server responds with an HTTP 400 (Bad Request)
   status code (unless specified otherwise) and includes the following
   parameters with the response:

   error
         REQUIRED.  A single ASCII error code from the
         following:

         invalid_request
               The request is missing a required parameter, includes an
               unsupported parameter value (other than grant type),
               repeats a parameter, includes multiple credentials,
               utilizes more than one mechanism for authenticating the
               client, or is otherwise malformed.

         invalid_client
               Client authentication failed (e.g., unknown client, no
               client authentication included, or unsupported
               authentication method).  The authorization server MAY
               return an HTTP 401 (Unauthorized) status code to indicate
               which HTTP authentication schemes are supported.  If the
               client attempted to authenticate via the "Authorization"
               request header field, the authorization server MUST
               respond with an HTTP 401 (Unauthorized) status code and
               include the "WWW-Authenticate" response header field
               matching the authentication scheme used by the client.

         invalid_grant
               The provided authorization grant (e.g., authorization
               code, resource owner credentials) or refresh token is
               invalid, expired, revoked, does not match the redirection
               URI used in the authorization request, or was issued to
               another client.

         unauthorized_client
               The authenticated client is not authorized to use this
               authorization grant type.

         unsupported_grant_type
               The authorization grant type is not supported by the
               authorization server.
```

So we expect it to be a 400 unless it is an invalid client.

Perhaps having this be a strict requirement wasn't the intent of the spec writers though (no must/shall on the 400), and maybe there is no reason to not allow other codes.  Are we interpreting 'unless otherwise specified' incorrectly maybe?

Are there other major implementations out there that do something other than 400?  I'm not sure that a single implementation defines 'industry standard', though I suspect Auth0 carries a lot of weight here.

yunwwang:
We had a similar discussion on v1.8

https://github.com/onc-healthit/inferno-program/issues/422

kayadanae:
Hey, thanks for your quick feedback! I don't know why Auth0 chose to diverge from the RFC. [This blog post](https://auth0.com/blog/forbidden-unauthorized-http-status-codes/) describes somewhat how Auth0 thinks about each of the three big 40X errors, but it's not particularly technical in nature. Would it be helpful for me to reach out to our Auth0 liaison to try to get more clarity into why their system operates this way?

My questions to you would be: Are there known compromises to interoperability or security being made if our application returns 403 instead of 400? And what value are we producing for our patients and collaborators if we insert this shim to comply with this requirement?

Thanks again!

arscan:
Hi @kayadanae --

It might be a good idea to reach out to the Auth0 folks to see what they think, if you have a way of doing that.  Perhaps they made a mistake here and are interested in fixing it?

I also see that there is an [OAuth 2.1 draft](https://oauth.net/2.1/) out there, but [the language for this section hasn't changed](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-05#section-3.2.3.1).  If Auth0 thinks that OAuth 2.0 is misleading or wrong, could they get this language changed in 2.1?  

I don't know how loosening this check would affect interop. As long as the system doesn't return a success status code then I doubt it would make a difference for clients.  But the spec authors put that line in for a reason, presumably.

I agree that a bad outcome would be a disagreement here that results in a bunch of shims.

My best argument for removing it is that there is no explicit SHALL/MUST there, and there is a fuzzy "(unless specified otherwise)" qualifier that isn't 100% clear about who is adding that specificity.  My main concern is that then we have no hard requirement for requiring any check on status there (why just allow 400, 403 -- why not others?).  And secondarily, there might be other important requirements where there is technically no SHALL/MUST statement, and will lose the ability to enforce other requirements that are important.

Rob

kayadanae:
Hey Rob,

Appreciate the thoughtfulness here.

I did reach out to Auth0 and unfortunately they had no documentation justifying their diversion from the spec, and in fact it’s apparently on their Q4 roadmap to correct this behavior and achieve conformity with the spec. Given that, I don’t think it makes sense for inferno to change the requirements. We’ll be inserting the shim, which will hopefully be temporary.

Thanks again,
Kaya

arscan:
Thanks for reaching out to Auth0 on this Kaya.  I'll go ahead and close this. I hope you don't have to maintain this workaround for long.

PR #85, opened 2022-04-11
Fix matrix generator so it doesnt break if no description on a test

arscan:
Noticed that we didn't have the regenerated matrix in yet, and was going to push it up to the release branch, but found a bug that is fixed here... which led me to morre bugs in us core which are waiting to get pushed through (which is why I didn't include the actual improved xls file in here)


arscan:
Addressed separately.

Issue #93, opened 2022-04-13
Single Patient API Test - Ruby Error without "Additional Patient Ids"


spclbp:
WIth the newly released ONC Certification (g)(10) Standardized API V.2.1.0, we've come upon a new error. First we perform the EHR Launch test to get our access token, then head over to the Single Patient API tests. We're getting the Ruby error below when we do not fill in the additional patient ids field:Error: 

undefined method `split' for nil:NilClass
/usr/local/bundle/gems/us_core_test_kit-0.2.1/lib/us_core_test_kit/search_test.rb:436:in `patient_id_list'

Work around:  Enter patient ids into the additional patient ids field -- even if it's the same id that's already present in the "Patient ID from SMART App Launch" field.

arscan:
Thanks for the report @spclbp -- we're working on a fix and will likely provide a patch release ASAP.

PR #94, opened 2022-04-13
FI-1533: Fix nil safety in additional_patient_ids

Jammjammjamm:
#93 describes a nil-safety issue with the `additional_patient_ids` input. This branch makes it nil safe.

### DIFF
--- a/lib/onc_certification_g10_test_kit/single_patient_api_group.rb
+++ b/lib/onc_certification_g10_test_kit/single_patient_api_group.rb
@@ -61,11 +61,15 @@ module ONCCertificationG10TestKit
       run do
         smart_app_launch_patient_id = patient_id.presence
         additional_patient_ids_list =
-          additional_patient_ids
-            .split(',')
-            .map(&:strip)
-            .map(&:presence)
-            .compact
+          if additional_patient_ids.present?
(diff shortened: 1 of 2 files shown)


lib/onc_certification_g10_test_kit/single_patient_api_group.rb
```diff
@@ -61,11 +61,15 @@ class SinglePatientAPIGroup < Inferno::TestGroup
...
+            additional_patient_ids
+              .split(',')
+              .map(&:strip)
+              .map(&:presence)
+              .compact
+          else
+            []
+          end
 
         all_patient_ids = ([smart_app_launch_patient_id] + additional_patient_ids_list).compact.uniq
```
yunwwang, line 74: This would create duplicates in the combine array:
[1] + [1, 2] => [1, 1, 2]

yunwwang: Never Mind. I saw you use .uniq at the end.

arscan (approved):
Verified that this now works when the Additional Patients field is blank, and when there is an entry in the Addiotnal Patient field too.

yunwwang (approved):

PR #95, opened 2022-04-13
Release v2.1.1

Jack-Fraser16:
Release patch for the bug caused by leaving the 'Additional Patient IDs' input blank in Single Patient API group.


arscan:
Could you also open up the test matrix xls and up the version in the top of the first sheet to v2.1.1 and commit it?  You don't have to regenerate it from scratch (there is a bug preventing it right now we need to fix in US Core), but I would like that version number to be accurate... so manually is fine.

arscan (approved):

PR #103, opened 2022-04-20
FI-1524 bulk validation

yunwwang:
This PR changes Bulk Data export validation so that

* it validates all lines/resources
* it saves validation errors and warnings from the first failed line/resource
* it returns result message with the count of failed lines/resources

This PR fix Github Issue #104

### DIFF
--- a/lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
+++ b/lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
@@ -169,9 +169,14 @@ module ONCCertificationG10TestKit
         resources[profile_url] << resource
         scratch[:patient_ids_seen] = patient_ids_seen | [resource.id] if resource_type == 'Patient'
 
-        if !resource_is_valid?(resource: resource, profile_url: profile_url) && !first_error.key?(:line_number)
-          first_error[:line_number] = line_count
-          first_error[:messages] = messages.select { |message| ['error', 'warning'].include? message[:type] }
+        unless resource_is_valid?(resource: resource, profile_url: profile_url)
+          if first_error.key?(:line_number)
+            @invalid_line_count += 1
+          else
+            @invalid_line_count = 1
+            first_error[:line_number] = line_count
+            first_error[:messages] = messages.select { |message| ['error', 'warning'].include? message[:type] }
+          end
         end
       }
 
@@ -190,16 +195,15 @@ module ONCCertificationG10TestKit
     end
 
     def process_validation_errors(total)
-      error_count = messages.count { |message| message[:type] == 'error' }
-      return if error_count.zero?
+      return if @invalid_line_count.zero?
 
       first_error_message = "The line number for the first failed resource is #{first_error[:line_number]}."
 
       messages.clear
       messages.concat(first_error[:messages])
 
-      assert error_count.zero?,
-             "#{error_count} / #{total} #{resource_type} resources failed profile validation. #{first_error_message}"
+      assert @invalid_line_count.zero?,
+             "#{@invalid_line_count} / #{total} #{resource_type} resources failed profile validation. #{first_error_message}"
     end
 
     def perform_bulk_export_validation
@@ -216,6 +220,7 @@ module ONCCertificationG10TestKit
 
       @resources_from_all_files = {}
       @first_error = {}
+      @invalid_line_count = 0
       success_count = 0
 
       file_list.each do |file|
--- a/lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
+++ b/lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
@@ -55,6 +55,10 @@ module ONCCertificationG10TestKit
       @resources_from_all_files ||= {}
     end
 
+    def first_error
+      @first_error ||= {}
+    end
+
     def patient_ids_seen
       scratch[:patient_ids_seen] ||= []
     end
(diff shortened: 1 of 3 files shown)


lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
```diff
@@ -197,11 +220,16 @@ def perform_bulk_export_validation
       end
 
       @resources_from_all_files = {}
+      @first_error = {}
+      @invalid_line_count = 0
```
Jammjammjamm, line 224: I'm not sure why these are initialized here and above.

spec/onc_certification_g10_test_kit/bulk_data_group_export_validation_spec.rb
```diff
@@ -150,6 +172,23 @@ def run(runnable, inputs = {})
 
       expect(result.result).to eq('pass')
     end
+
+    it 'validates all lines and saves errors for the first failed line' do
+      stub_request(:get, endpoint.to_s)
+        .with(headers: { 'Accept' => 'application/fhir+ndjson' })
+        .to_return(status: 200, body: contents, headers: headers)
+
+      stub_request(:post, "#{validator_url}/validate?profile=http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient")
```
Jammjammjamm, line 181: We expect this request to happen twice, so as a sanity check we should make sure that actually happens.

```
validation_request = stub_request(...)

...the rest of the spec...

expect(validation_request).to have_been_made.twice
```

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
```diff
@@ -197,11 +220,16 @@ def perform_bulk_export_validation
...
       success_count = 0
+
       file_list.each do |file|
         success_count += check_file_request(file['url'])
       end
 
+      process_validation_errors(success_count)
```
Jammjammjamm, line 231: It's confusing to me that `success_count` is being passed as an argument to `process_validation_errors`, when it seems that inside `process_validation_errors` it represents the total number of resources. It may be that `success_count` made sense as a variable name before, but now that we're continuing on validation failures a name like `resource_count` would make more sense now.

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
```diff
@@ -166,7 +170,13 @@ def check_file_request(url) # rubocop:disable Metrics/CyclomaticComplexity
         scratch[:patient_ids_seen] = patient_ids_seen | [resource.id] if resource_type == 'Patient'
 
         unless resource_is_valid?(resource: resource, profile_url: profile_url)
-          assert false, "Resource at line \"#{line_count}\" does not conform to profile \"#{profile_url}\"."
+          if first_error.key?(:line_number)
+            @invalid_line_count += 1
```
Jammjammjamm, line 174: `@invalid_resource_count` would make more sense to me.

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
```diff
@@ -166,7 +170,13 @@ def check_file_request(url) # rubocop:disable Metrics/CyclomaticComplexity
...
+          else
+            @invalid_line_count = 1
+            first_error[:line_number] = line_count
+            first_error[:messages] = messages.select { |message| ['error', 'warning'].include? message[:type] }
```
Jammjammjamm, line 178: I don't think you should need to filter the messages here. `resource_is_valid?` filters them based on however the validator was configured to filter them.

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
```diff
@@ -184,6 +194,19 @@ def check_file_request(url) # rubocop:disable Metrics/CyclomaticComplexity
       line_count
     end
 
+    def process_validation_errors(total)
+      return if @invalid_line_count.zero?
+
+      first_error_message = "The line number for the first failed resource is #{first_error[:line_number]}."
+
+      messages.clear
+      messages.concat(first_error[:messages])
```
Jammjammjamm, line 203: This seems unnecessary given the assertion below.

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
yunwwang: Haha. I had the name `@invalid_resource_count ` but changed to `@invalid_line_count` in the last minute. I will change it back

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
yunwwang: This is to keep errors from the first failed resource and pass to the test_result and repository later

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
yunwwang: Changed to message.dup

lib/onc_certification_g10_test_kit/bulk_export_validation_tester.rb
Jammjammjamm: Oh, right. I was mixing up `first_error[:messages]` and `first_error_message`.

Jammjammjamm (approved):

PR #107, opened 2022-04-25
FI-1546: Dev versioning

Jammjammjamm:
This branch adds a configuration error message when using a development version. When not using an official release, the version will end in `.dev`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    onc_certification_g10_test_kit (2.1.1)
+    onc_certification_g10_test_kit (2.1.1.dev)
       bloomer (~> 1.0.0)
       colorize (~> 0.8.1)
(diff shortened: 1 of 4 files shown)


yunwwang (approved):

arscan (approved):
Looks great.  Do we have documentation on how to add configuration/installation validation to a test suite?  We should add a ticket for that.  I looked and didn't see anything obvious.

My guess is that we have documentation for adding it to a test suite via inferno-core through the regular gem docs, but not in a test kit generated via the template through some simple hooks.

Jammjammjamm:
> Do we have documentation on how to add configuration/installation validation to a test suite?

It's just in the gem docs. I'll make a ticket, but I think this is advanced/obscure enough functionality that there are plenty of other things that could use better documentation first.

I feel like we shouldn't add it to the template for similar reasons.

PR #118, opened 2022-05-06
FI-1556 Updates links in descriptions to correct version of Bulk IG per regs.

arscan:
The regulation is technically to v1.0.1 of Bulk Data, which has slightly different links than v1.0.0.  This was raised to ONC via their comment portal.  Technically the changes are very small and do not affect the parts that we linked to generally, but we need to be very precise.

I also found a link to build.fhir.org separately which is incorrect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ supports all required standards, including:
 * Health Level 7 (HL7®) Fast Healthcare Interoperability Resources (FHIR®) Release 4.0.1
 * FHIR US Core Implementation Guide (IG) STU 3.1.1
 * SMART Application Launch Framework Implementation Guide Release 1.0.0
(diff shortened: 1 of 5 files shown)


Jack-Fraser16 (commented):
In the first group export test, where the CapabilityStatement is checked for the export operation, should the following url be used in either the checking or the description: https://hl7.org/fhir/uv/bulkdata/STU1.0.1/OperationDefinition-group-export.html? If not, why does the versioning not matter here?

Jack-Fraser16 (approved):

arscan:
Also, i didn't mention why we just strip the version here -- its probably a good practice to allow systems that point to a later version to still pass by default.  We would have to make a rock-solid case if we decided to only allow a specific version of the operation definition, and we do not have that.

PR #122, opened 2022-05-10
Make Version .dev

Jack-Fraser16:
Ensures users will receive error message if using anything but official release version


radamson (approved):
I see the expected configuration error warning users that this is a development version not suitable for certification.

PR #150, opened 2022-06-30
FI-1582: Bulk Data STU2 _outputFormat

Jack-Fraser16:
# Summary
Reorganizes the Multi-Patient Api test group into two separate groups (STU1 and STU2). Both are included and shown within the G10 suite, but this will be updated once inferno-core SVAP changes are incorporated. Added a test to the STU2 group that checks whether different content types for the 'outputFormat' query parameter are supported by the endpoint. Also, updated the presets to reflect bulk data server being merged into reference server.

# Testing this PR
Test with Docker using:
```shell
./setup.sh
./run.sh
```
Note that the `outputFormat` test added to Multi-Patient API STU2 fails as the reference server has not been updated to handle this query parameter.


lib/onc_certification_g10_test_kit.rb
```diff
@@ -140,6 +141,8 @@ def self.well_known_route_handler
 
     group from: 'multi_patient_api'
 
+    group from: 'multi_patient_api_stu2'
+
     group do
```
Jammjammjamm, line 146: This group needs to be given an id of `Group06` to keep its id.

lib/onc_certification_g10_test_kit/multi_patient_api_stu2.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      Register Inferno with the following JWK Set Url:
+
+      * `#{Inferno::Application[:base_url]}/custom/g10_certification/.well-known/jwks.json`
+    )
+
+    description %(
+      Demonstrate the ability to export clinical data for multiple patients in
+      a group using [FHIR Bulk Data Access
+      IG](https://hl7.org/fhir/uv/bulkdata/). This test uses [Backend Services
```
Jammjammjamm, line 24: These links should all point to specific IG versions.

lib/onc_certification_g10_test_kit/bulk_data_group_export_stu2.rb
```diff
@@ -0,0 +1,30 @@
...
+    description <<~DESCRIPTION
+      Verify that system level export on the Bulk Data server follow the Bulk Data Access Implementation Guide
+    DESCRIPTION
+
+    id :bulk_data_group_export_stu2
+
+    test do
+      title 'Bulk Data Server supports "_outputFormat" query parameter'
+      description <<~DESCRIPTION
+        _outputFormat: The format for the requested Bulk Data files to be generated as per FHIR Asynchronous Request Pattern. Defaults to application/fhir+ndjson. The server SHALL support Newline Delimited JSON, but MAY choose to support additional output formats. The server SHALL accept the full content type of application/fhir+ndjson as well as the abbreviated representations application/ndjson and ndjson.
```
Jammjammjamm, line 17: This line needs to be wrapped, and it would be nice to link to the relevant part of the IG here.

Jammjammjamm (commented):
Should we be concerned about kicking off three exports in quick succession that won't actually be used? Should we cancel the requests?

lib/onc_certification_g10_test_kit/bulk_data_group_export_stu2.rb
Jack-Fraser16: Done.

Jack-Fraser16:
> Should we be concerned about kicking off three exports in quick succession that won't actually be used? Should we cancel the requests?

I implemented a change that uses the same cancellation method as the final test in the group -- grab the polling url and send the cancellation request to there. Added a method to do this and incorporated it into the tests.

Jack-Fraser16:
> Should we be concerned about kicking off three exports in quick succession that won't actually be used? Should we cancel the requests?

I implemented a change that uses the same cancellation method as the final test in the group -- grab the polling url and send the cancellation request to there. Added a method to do this and incorporated it into the tests.

lib/onc_certification_g10_test_kit/multi_patient_api_stu2.rb
Jammjammjamm: If you want to update the descriptions of the children, yes, each one needs to be imported and updated indivually.

lib/onc_certification_g10_test_kit.rb
```diff
@@ -140,6 +141,9 @@ def self.well_known_route_handler
 
     group from: 'multi_patient_api'
 
+    group from: 'multi_patient_api_stu2',
+          id: 'Group06'
```
Jammjammjamm, line 145: This is on the wrong group. It got added in another PR that was merged, so you can just remove it.

lib/onc_certification_g10_test_kit/bulk_data_group_export_stu2.rb
```diff
@@ -0,0 +1,38 @@
...
+    description <<~DESCRIPTION
+      Verify that system level export on the Bulk Data server follow the Bulk Data Access Implementation Guide
+    DESCRIPTION
+
+    id :bulk_data_group_export_stu2
+
+    test do
+      title 'Bulk Data Server supports "_outputFormat" query parameter'
+      description <<~DESCRIPTION
+        [_outputFormat](https://hl7.org/fhir/uv/bulkdata/export.html#query-parameters):
```
Jammjammjamm, line 17:
```suggestion
        [_outputFormat](https://hl7.org/fhir/uv/bulkdata/STU2/export.html#query-parameters):
```

lib/onc_certification_g10_test_kit.rb
```diff
@@ -140,6 +141,9 @@ def self.well_known_route_handler
 
     group from: 'multi_patient_api'
 
+    group from: 'multi_patient_api_stu2',
```
Jammjammjamm, line 144: The SVAP stuff is all in, as are the feature flags, so this should be updated to use them.

lib/onc_certification_g10_test_kit/bulk_data_group_export_stu2.rb
```diff
@@ -0,0 +1,52 @@
...
+  class BulkDataGroupExportSTU2 < Inferno::TestGroup
+    title 'Group Compartment Export Tests STU2'
+    short_description 'Verify that the system supports Group compartment export.'
+    description <<~DESCRIPTION
+      Verify that system level export on the Bulk Data server follow the Bulk Data Access Implementation Guide
+    DESCRIPTION
+
+    id :bulk_data_group_export_stu2
+
+    test from: 'bulk_data_group_export-g10_bulk_data_server_tls_version'
```
Jack-Fraser16, line 14: Pulling in each individual test for this Group as additional Bulk Data Group Export STU2 changes (beyond this PR, whose `_outputFormat` test is additive only) will alter specific tests. This is not the case for  Bulk Data Group Export STU1, which inherits directly from Bulk Data Group Export TestGroup.

lib/onc_certification_g10_test_kit/bulk_data_group_export.rb
```diff
@@ -79,9 +73,10 @@ class BulkDataGroupExport < Inferno::TestGroup
         Additionally, this test provides a warning if the bulk data server does
```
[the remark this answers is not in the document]
Jack-Fraser16: My reasoning is that we're not actually using the URL to fail them, so further specifying might not be useful and would necessitate altering the test between the different versions.

Jack-Fraser16:
It is very cool how easy it was to make this multi-version @Jammjammjamm

Screen Shot 2022-07-15 at 12 47 53 PM

I think it's probably too early to settle on naming conventions (like I don't know if I love having the specific version be included in sub-group names). But I think this does the trick for now.

Screen Shot 2022-07-15 at 12 48 02 PM

To test, set `ENABLE_BULK_DATA_V2=true`

Jammjammjamm:
> I think it's probably too early to settle on naming conventions (like I don't know if I love having the specific version be included in sub-group names)

Yeah, the nice thing about having it behind a feature flag is we can get it in, mess around with it, and then see what makes sense as more things get added. For now I think it's nice to have the version there (at least for STU2) so that we can quickly verify that we are in fact seeing the right tests.

lib/onc_certification_g10_test_kit/multi_patient_api_stu1.rb
```diff
@@ -33,7 +33,7 @@ class MultiPatientAPIGroup < Inferno::TestGroup
       Location, Organization, and Practitioner resources as they are
       referenced as must support elements in required resources.
     )
-    id :multi_patient_api
+    id :multi_patient_api_stu1
```
Jammjammjamm, line 36: These id changes break backwards compatiibility.

lib/onc_certification_g10_test_kit/bulk_data_group_export_stu1.rb
```diff
@@ -0,0 +1,8 @@
+require_relative 'bulk_data_group_export'
+
+module ONCCertificationG10TestKit
+  class BulkDataGroupExportSTU1 < BulkDataGroupExport
+    title 'Group Compartment Export Tests STU1'
+    id :bulk_data_group_export_stu1
```
Jammjammjamm, line 6: This is changing the id of this group. I don't think it's worth making a new class just to change the title.

lib/onc_certification_g10_test_kit/multi_patient_api_stu1.rb
```diff
@@ -47,8 +47,22 @@ class MultiPatientAPIGroup < Inferno::TestGroup
...
+
+            In this set of tests, Inferno serves as a Bulk Data client that requests authorization
+            from the Bulk Data authorization server.  It also performs a number of negative tests
+            to validate that the authorization service does not improperly authorize invalid
+            requests.
+
+            This test returns an access token.
+          DESCRIPTION
+
+    group from: :bulk_data_group_export_stu1
```
Jammjammjamm, line 65: This is where the id changes and will cause problems.

Jammjammjamm (approved):

PR #151, opened 2022-06-30
FI-1474: Improve wait messages

Jammjammjamm:
This branch improves the g10 suite's wait messages. WIP because this branch will require:
* a new core release which includes https://github.com/inferno-framework/inferno-core/pull/211
* a new SMART test kit release which includes https://github.com/inferno-framework/smart-app-launch-test-kit/pull/25

Screenshots of the updated wait messages:
Screen Shot 2022-06-30 at 9 03 37 AM
Screen Shot 2022-06-30 at 9 04 00 AM
Screen Shot 2022-06-30 at 9 04 29 AM
Screen Shot 2022-06-30 at 9 04 56 AM
Screen Shot 2022-06-30 at 9 07 44 AM
Screen Shot 2022-06-30 at 9 09 30 AM

### DIFF
--- a/lib/onc_certification_g10_test_kit.rb
+++ b/lib/onc_certification_g10_test_kit.rb
@@ -178,6 +178,22 @@ module ONCCertificationG10TestKit
         may require special setup on the part of the tester.
       )
 
+      config(
+        options: {
+          redirect_message_proc: lambda do |auth_url|
+            %(
+              ### #{self.class.parent.title}
+
+              [Follow this link to authorize with the SMART
+              server](#{auth_url}).
+
+              Tests will resume once Inferno receives a request at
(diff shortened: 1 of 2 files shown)


arscan (approved):
Looks good.  I do think it might be helpful in the EHR one to have something a bit more specific about what the user is expected to do, instead of the fact that Inferno is waiting.

E.g. instead of 'Waiting for Inferno to be launched'... say 'Inferno is ready to be launched.  Please initiate a launch to the registered Inferno application from within the EHR.'

Looks like that isn't part of this PR though -- if you think the language is an improvement feel free to update wherever necessary while we are looking at this.  If not, no big deal.

Jammjammjamm:
Yeah, the EHR launch message isn't changed here, so we can update that in the SMART test kit separately.

Issue #98, opened 2022-04-18
Invalid launch parameter tests exceed spec requirements


Jammjammjamm:
The _SMART Invalid Launch Parameter_ tests (6.4) require that a server fail to authorize a client which sends an invalid `launch` parameter. The SMART App Launch IG [states that clients send back the same `launch` parameter they received from the server,](http://hl7.org/fhir/smart-app-launch/1.0.0/#step-1-app-asks-for-authorization) but does not state any requirements for how the server handles an invalid or missing `launch` parameter.

See [discussion on zulip.](https://chat.fhir.org/#narrow/stream/179170-smart/topic/Server.20response.20when.20client.20sends.20bad.20launch.20parameter)

alps8:
@yunwwang @arscan 
Hi,
  We are currently using [v2.2.1] and this test case is failing for us. Is there an ETA or version number where this constraint will be removed? Thanks.

PR #152, opened 2022-07-01
FI-1615: Remove invalid launch tests

Jammjammjamm:
This branch removes the invalid `launch` parameter tests, addressing #98.


arscan (changes requested):
If this changes numbering you'll have to update https://github.com/onc-healthit/onc-certification-g10-test-kit/blob/main/lib/onc_certification_g10_test_kit/onc_program_procedure.yml

We should also review the TP to see how ONC removed this (did their numbering change too, or is this just a wording change to a step instead of removal?)

We used to have some kind of 'deprecated' flag that retains numbering but hides the test. Not sure if its worth it though -- keeping dead code (or better, an empty test/group that is just more-or-less a placeholder of what was once a test) around doesn't seem managable long-term.

arscan (approved):
I updated the procedure to include the new language from ONC's site, and regenerated the matrix again.

I created a session on `main`, ran the test, then switched to this branch and it doesn't look like the system has a problem with old session with that extra data.  Looks good.

PR #162, opened 2022-07-08
Release v2.2.2

Jack-Fraser16:


arscan (approved):
I don't see anything wrong.  Do we do a prerelease gem real quick to try it out through the inferno-deployment repo and on `inferno-dev.healthit.gov`?  Or do we just release it like this and then check it going through inferno-deployment after?

PR #163, opened 2022-07-11
FI-1621: Add feature flags

Jammjammjamm:
This branch adds support for basic feature flags to enable/disable work in progress features. Set `US_CORE_4_ENABLED` to `true` to see a version of the g10 suite with US Core 4 support.

Is currently WIP because it needs a new core release.


chris-at-work (approved):
Well this PR/diff mostly makes sense to me as it stands now.  Approving so WIP and dependency on core can be done at your discretion.

Issue #172, opened 2022-07-15
Allow for custom bulk data JWKS


nathanloyer:
I'm currently working on a new iteration of my company's internal customization of inferno. Our first iteration was a fork of this repo with some small customizations and creating some of our own test suites. Now I'm trying to convert it to being it's own test kit. One thing we do in our deployment is load in our own file to replace the file `lib/onc_certification_g10_test_kit/bulk_data_jwks.json` in our docker image. We do this because we want the additional security of only allowing our fhir server to accept requests coming from our deployment of inferno and not allow the public inferno to access it. 

This feature request is to allow us to specify a file to use for the JWKS, so we don't have to copy the file into the ruby gems directory over the file that you package with the gem. I think setting this via environment variable or one of the config files would be sufficient. Or any other solution as long as it is easy for us to configure on our own.

PR #173, opened 2022-07-15
Allow for custom bulk data JWKS

nathanloyer:
add the G10_BULK_DATA_JWKS env var to customize the JWKS used for bulk tests

see more info on #172

### DIFF
--- a/.env
+++ b/.env
@@ -4,7 +4,7 @@ G10_VALIDATOR_URL=http://validator_service:4567
 REDIS_URL=redis://redis:6379/0
 
 # Full path to a custom JWKS json file, will use lib/onc_certification_g10_test_kit/bulk_data_jwks.json if left blank
-G10_BULK_DATA_JWKS=
+# G10_BULK_DATA_JWKS=
 
 # Flags for features under development
 US_CORE_4_ENABLED=false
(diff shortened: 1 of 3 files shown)


nathanloyer:
Thanks for the review @Jammjammjamm, lmk if there's anything else

nathanloyer:
I see the test error, wont have time to dig into the cause for a couple of days

.env
```diff
@@ -3,6 +3,9 @@ INFERNO_HOST=http://localhost
 G10_VALIDATOR_URL=http://validator_service:4567
 REDIS_URL=redis://redis:6379/0
 
+# Full path to a custom JWKS json file, will use lib/onc_certification_g10_test_kit/bulk_data_jwks.json if left blank
+G10_BULK_DATA_JWKS=
```
Jammjammjamm, line 7:
```suggestion
# G10_BULK_DATA_JWKS=
```

nathanloyer: oh you meant for it to be commented it, I thought you wanted it to be set. got it

nathanloyer:
looks like that was the cause of the test failure too

Jammjammjamm (approved):

Issue #170, opened 2022-07-14
Modified scope is not carried forward to the appropriate section (FI-1639)


IgLysk00:
Inferno defaults the following scope on the Standalone Patient App - Full Access input screen:
 launch/patient openid fhirUser offline_access patient/*.read
The first issue is that not all FHIR resources can use patient/*.read, for example scope for the  Organization resource should be user/Organization.read. But, the most important is that when scope is modified it is not carried forward to the next appropriate series of tests. Here is an example. 
Using Inferno Reference Server modify scope on the Standalone Patient App - Full Access input screen as follows:

launch/patient openid fhirUser offline_access patient/AllergyIntolerance.read patient/CarePlan.read patient/CareTeam.read patient/Condition.read patient/Device.read patient/DiagnosticReport.read patient/DocumentReference.read patient/Encounter.read patient/Goal.read patient/Immunization.read patient/MedicationRequest.read patient/Observation.read user/Organization.read patient/Patient.read user/Practitioner.read patient/Procedure.read patient/Provenance.read user/Person.read
Run the Standalone Patient App - Full Access series of tests. Navigate to Standalone Patient App - Limited Access. Select Run Tests and observe that the default scope values are displayed instead of the modified scope entered before.

arscan:
Thanks for logging a ticket for the suggestion to carry forward the scopes to the Limited Access test.  I believe that prior to v2.0 we did it in the way you are suggesting.  There may be a technical reason why we decided to not do that in v2.0+, but it could just be an oversight.  We are looking into it.

Regarding the other point:

> The first issue is that not all FHIR resources can use patient/\*.read, for example scope for the Organization resource should be user/Organization.read. 

This is an implementation decision to be made by the server.  `patient/Organization.read` could grant access to those Organizations needed to make sense of the patient's compartment, even if though the Organization resource does not have a direct reference to the patient (a couple of hops would be involved). We defaulted to this approach because we are testing the [permission-patient](http://hl7.org/fhir/smart-app-launch/1.0.0/conformance/index.html#permissions) core capability in this scenario, but we allow systems to change that to user/Organization.read if they have chosen the more literal approach where patient/\*.read scopes only apply to resources that have a direct patient reference in them.

There have been a couple of conversations about this on Zulip (it is certainly a point of confusion) -- one example is here: https://chat.fhir.org/#narrow/stream/179170-smart/topic/patient.2F*.2E*

Does this clear up that issue for you?

[...]

PR #183, opened 2022-07-26
FI-1639 Update limited app launch to have launch use same inputs as in Standalone App Launch test

arscan:
**Note for release notes: this removes an input so presets should be updated.  A non-fatal warning will be issued in the console otherwise.**

This updates the input to the Limited App Launch test to use the same inputs as in the Standalone App Launch test because the purpose of the test to ensure that the user has control over which scopes are granted, and that an app doesn't have to request different scopes (which the user likely will not have control over).  This makes this test consistent with the behavior in <v1.9, as it was inadvertently changed.

To test, verify that the limited launch still works properly, and that whatever scopes and pkce-related fields that are used in the Standalone App Launch test are persisted in this input.  Also, the pkce-related fields will now be locked.

Screen Shot 2022-07-27 at 12 29 05 PM

See #170

### DIFF
--- a/lib/onc_certification_g10_test_kit/smart_limited_app_group.rb
+++ b/lib/onc_certification_g10_test_kit/smart_limited_app_group.rb
@@ -8,11 +8,14 @@ module ONCCertificationG10TestKit
     short_title 'Limited Access App'
 
     input_instructions %(
-      The purpose of this test is to demonstrate that users can restrict access
-      granted to apps to a limited number of resources. Enter which resources the
-      user will grant access to below, and during the launch process only grant
-      access to those resources. Inferno will verify that access granted matches
-      these expectations.
+      The purpose of this test is to demonstrate that app users can restrict
+      access granted to apps to a limited number of resources. Enter which
+      resources the user will grant access to below, and during the launch
+      process only grant access to those resources. Inferno will verify that
+      access granted matches these expectations.
+
+      All other inputs are locked to ensure the same app configuration as in the
+      Standalone Patient App - Full Access test.
     )
 
     description %(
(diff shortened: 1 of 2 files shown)


lib/onc_certification_g10_test_kit/smart_limited_app_group.rb
```diff
@@ -67,14 +79,13 @@ class SmartLimitedAppGroup < Inferno::TestGroup
...
           url: { locked: true },
+          requested_scopes: { locked: true },
+          use_pkce: { locked: true },
+          pkce_code_challenge_method: { locked: true },
           code: { name: :limited_code },
           state: { name: :limited_state },
           patient_id: { name: :limited_patient_id },
           access_token: { name: :limited_access_token },
-          requested_scopes: {
-            name: :limited_requested_scopes,
```
Jammjammjamm, line 75: You can remove this from the preset. This input name also needs to be changed in lines 32, 155, and 166.

arscan:
Note: I created a separate internal ticket to make it so that locked radio buttons aren't a lighter grey than other locked inputs (FI-1651).

lib/onc_certification_g10_test_kit/smart_limited_app_group.rb
arscan: Good point, not sure why I didn't run a global search on this.  Updated.  I also noticed that client_secret isn't marked as 'required', which it should be. Wouldn't make a difference but might as well have it be consistent.

Jammjammjamm:
> We should note that any downstream users that have presets should update their presets as well, else they would receive a warning

Are you sure about that?

arscan:
> Are you sure about that?

Actually, no.  I guess I was thinking of when people copy/paste the fields through the UI.  Regardless, people should be informed that any presets should be changed, if only to remove values that are no longer relevant.

lib/onc_certification_g10_test_kit/smart_limited_app_group.rb
arscan: @Jammjammjamm I believe I addressed this, could you take another look?

Jammjammjamm (approved):

PR #186, opened 2022-07-28
FI-1649: Update terminology build for US Core 4

Jammjammjamm:
This branch updates the terminology build to include US Core 4.0.0 terminology. The main changes involve adding US Core 4 and VSAC (since it's used by US Core 4) to the list of FHIR packages which will be downloaded, and adding some special handling to avoid processing every VS in VSAC.


Jack-Fraser16:
How would I go about testing this?

Jammjammjamm:
Run the build. Check that it built correctly. Make sure that the changes to included terminology make sense and match what's in US Core 4.

arscan (approved):

PR #195, opened 2022-08-04
FI-1632: Update terminology for US Core 5

Jammjammjamm:


resources/value_sets.yml
```diff
@@ -1717,3 +1717,121 @@
...
+  :system: http://hl7.org/fhir/ValueSet/medication-as-needed-reason
+  :path: asNeeded
+- :type: CodeableConcept
+  :strength: example
+  :system: http://hl7.org/fhir/ValueSet/participant-role
+  :path: performerType
+- :type: CodeableConcept
+  :strength: example
+  :system: http://terminology.hl7.org/ValueSet/v3-ServiceDeliveryLocationRoleType
+  :path: locationCode
```
arscan, line 1837: I assume whatever context this is being used in knows what Resource type is relevant so you don't need to technically print out the Resource type here?  Might be nice to just put it in here if you have it at generation time so this is a more useful human-readable artifact for reviews too.  Just an idea -- I might be misinterpreting what this file is for.

Jammjammjamm: This file is used to check that all of the value sets we expect to be loaded are loaded. Adding the resource type would be easy if the file weren't made by hand.

lib/inferno/terminology/expected_manifest.yml
```diff
@@ -1182,13 +1320,28 @@
...
   :file: terminology_hl7_org_CodeSystem_medicationrequest-category.msgpack
   :count: 4
   :type: bloom
   :code_systems: http://terminology.hl7.org/CodeSystem/medicationrequest-category
-- :url: urn:oid:2.16.840.1.113883.6.101
-  :file: urn_oid_2_16_840_1_113883_6_101.msgpack
-  :count: 970
+- :url: http://www.nubc.org/patient-discharge
+  :file: www_nubc_org_patient-discharge.msgpack
+  :count: 0
```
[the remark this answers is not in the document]
Jammjammjamm: This is the new system with licensing restrictions, so we're not actually loading it. It isn't used in a required binding, so we don't actually validate against it. This file doesn't actually exist, though, so this entry is misleading. We may need a special case to handle this.

lib/inferno/terminology/expected_manifest.yml
```diff
@@ -678,15 +696,23 @@
...
 - :url: http://cts.nlm.nih.gov/fhir/ValueSet/2.16.840.1.113762.1.4.1010.6
   :file: cts_nlm_nih_gov_fhir_ValueSet_2_16_840_1_113762_1_4_1010_6.msgpack
-  :count: 172
+  :count: 196
   :type: bloom
   :code_systems:
   - http://hl7.org/fhir/sid/cvx
+- :url: http://cts.nlm.nih.gov/fhir/ValueSet/2.16.840.1.113762.1.4.1010.4
+  :file: cts_nlm_nih_gov_fhir_ValueSet_2_16_840_1_113762_1_4_1010_4.msgpack
+  :count: 1000
```
arscan, line 712: This is an oddly round number, not a cap?

resources/value_sets.yml
arscan: Oh, this file is made by hand?  Why do we need `path` then?

Jammjammjamm: It just comes straight from the US Core metadata: https://github.com/inferno-framework/us-core-test-kit/blob/main/lib/us_core_test_kit/generated/v3.1.1/metadata.yml#L93

lib/inferno/terminology/expected_manifest.yml
Jammjammjamm: It's not a cap. We have plenty of larger value sets.

lib/inferno/terminology/expected_manifest.yml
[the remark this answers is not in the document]
Jammjammjamm: I made an update so that this won't appear in the manifest.

arscan (approved):

PR #198, opened 2022-08-05
Remove standards version from test titles and fix typo.

arscan:
For this release, I think it is better to not have the standards version numbers in the test titles.  Part of the issue is that while convenient for us to verify that the right code is being loaded, it isn't quite 100% accurate because many of our tests actually cut across multiple standards.  Also, for this release, since most users won't use the SVAP versions of standards (requires downloading and changing feature flags), displaying version numbers is premature.

Screen Shot 2022-08-05 at 1 29 27 PM

This update just changes tests 4 and 5 to not state (US Core v3.1.1 and STU1), and fixes a small typo in another test description.

### DIFF
--- a/lib/onc_certification_g10_test_kit/multi_patient_api_stu1.rb
+++ b/lib/onc_certification_g10_test_kit/multi_patient_api_stu1.rb
@@ -4,8 +4,8 @@ require_relative 'bulk_data_group_export_validation'
 
 module ONCCertificationG10TestKit
   class MultiPatientAPIGroupSTU1 < Inferno::TestGroup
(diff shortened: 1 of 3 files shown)


Jack-Fraser16 (approved):

Jammjammjamm (commented):
You want to leave the versions in the titles of the groups for the new standards?

arscan:
> You want to leave the versions in the titles of the groups for the new standards?

I figure for this month its fine while we continue to refine this, since those will only be visible behind a feature flag.  I just wanted the regular g10 to look unchanged for right now.

PR #199, opened 2022-08-08
FI-1637: Include profile version in bulk validation

Jammjammjamm:
This branch updates the bulk validation tests to include the profile version in their validation calls.


Jack-Fraser16 (approved):

Jack-Fraser16:
Assuming the MedicationRequest/Observation validation skip/failure have to do with the resources not meeting USCore v4? Similar skip/ when running single-patient

Screen Shot 2022-08-09 at 4 42 19 PM

Jammjammjamm:
Yes, those are both related to US Core changes.

PR #200, opened 2022-08-08
FI-1581: Check that output file array contains absolute urls 

Jack-Fraser16:
Implementing [this ticket.](https://oncprojectracking.healthit.gov/support/secure/RapidBoard.jspa?rapidView=40&projectKey=FI&view=detail&selectedIssue=FI-1581&quickFilter=183)

### DIFF
--- a/lib/onc_certification_g10_test_kit/bulk_data_group_export_stu1.rb
+++ b/lib/onc_certification_g10_test_kit/bulk_data_group_export_stu1.rb
@@ -244,8 +244,7 @@ module ONCCertificationG10TestKit
 
         Each file SHALL contain resources of only one type, but a server MAY create more than one file for each resource type returned.
 
-        * url - the #{'absolute' if config.options[:require_absolute_urls_in_output]} path to the file.
-        The format of the file SHOULD reflect that requested in the _outputFormat parameter of the initial kick-off request.
+        * url - the path to the file. The format of the file SHOULD reflect that requested in the _outputFormat parameter of the initial kick-off request.
       DESCRIPTION
       # link 'http://hl7.org/fhir/uv/bulkdata/STU1.0.1/export/index.html#response---complete-status'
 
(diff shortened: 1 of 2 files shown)


lib/onc_certification_g10_test_kit/bulk_data_group_export_stu1.rb
```diff
@@ -244,7 +244,8 @@ class BulkDataGroupExportSTU1 < Inferno::TestGroup
 
         Each file SHALL contain resources of only one type, but a server MAY create more than one file for each resource type returned.
 
-        * url - the path to the file. The format of the file SHOULD reflect that requested in the _outputFormat parameter of the initial kick-off request.
+        * url - the #{'absolute' if config.options[:require_absolute_urls_in_output]} path to the file.
```
Jack-Fraser16, line 247:
```suggestion
        * url - the #{'absolute' if config.options[:require_absolute_urls_in_output]} path to the file.
```
This string interpolation isn't working as intended -- seems _config.options_ is an empty hash at the time of interpolation. If there's a quick alternative, let me know.

Jammjammjamm: Descriptions are evaluated at boot time, and outside of the context of a test session, so suite options are unavailable. Hacking in a workaround doesn't seem worth it for such a minor change.

lib/onc_certification_g10_test_kit/bulk_data_group_export_stu1.rb
```diff
@@ -265,6 +265,10 @@ class BulkDataGroupExportSTU1 < Inferno::TestGroup
           ['type', 'url'].each do |key|
             assert file.key?(key), "Output file did not contain \"#{key}\" as required"
           end
+
+          if config.options[:require_absolute_urls_in_output]
+            assert file['url'].match?(%r{\Ahttps?://}), 'URLs in output file were not absolute as required.'
```
Jammjammjamm, line 270: I think this error message should be something like:
```
"URLs in output file must be absolute, but found `#{file['url']}`"
```
Also, we've verified that the `url` key is present, but not that its value is a String, so it isn't safe to call `match` on it.

Jack-Fraser16: Added `to_s` which is defined on each data type, so should be fine?

Jammjammjamm (approved):

arscan:
I'm going to go ahead and merge @Jack-Fraser16 because it is approved.

PR #201, opened 2022-08-08
Release v2.3.0

arscan:


Jammjammjamm (commented):
There's this change as well: https://github.com/inferno-framework/us-core-test-kit/pull/56

arscan:
> There's this change as well: https://github.com/inferno-framework/us-core-test-kit/pull/56

I described this in the [Release Notes](https://github.com/onc-healthit/onc-certification-g10-test-kit/releases/tag/untagged-9bca6f6f15d6a2b7c7f4).

I'm confused what is supposed to be in the CHANGELOG vs. Release Notes -- I basically had the CHANGELOG be a 1-1 representation of PRs into this repo (e.g. code changed here), whereas the Release Notes is a human-readable rollup of all tester-visible changes that affect certification (test changes here or in any sub-test kit, as well as UI/UX updates in core).

Are we instead having the CHANGELOG roll up all changes in other test kits?  What about other libraries?

Jammjammjamm:
I included them the last time I did release notes (2.1.0), but maybe it hasn't been done since then.

Jammjammjamm (approved):

arscan:
> I included them the last time I did release notes (2.1.0), but maybe it hasn't been done since then.

Oh, I wasn't aware of this -- I see that now.  I don't understand the value of this particular file, since its redundant to what git / GitHub provides and the release notes, so I'd prefer we keep effort associated with this as low as possible.  Can I just state that we updated to new versions of those test kits / inferno-core instead of copy/pasting all updates to those here too?  Or should we just keep a mirror between the 'release notes' and the 'changelog' content instead of restating the same thing for different audiences?

Jammjammjamm:
I like having a plain text file in the repo because it's faster to look through that in my editor than to look through the github releases, but that isn't exactly something I do often, so I really don't care.

PR #211, opened 2022-08-12
FI-1675: US Core 5

Jammjammjamm:
This branch adds US Core 5 tests to the single patient api. It's just pointing at `main` on the US Core test kit rather than an official release because I don't think we want to slow down g10 development waiting on US Core releases.


yunwwang:
You can disable test 8.5.11 and 8.5.12 for uscore-4 and uscore-5 because patient's suffix and previous name MS tests are included in patient test group (5.2.09 and 6.2.09) for uscore-4 and uscore-5

Jammjammjamm:
> You can disable test 8.5.11 and 8.5.12 for uscore-4 and uscore-5 because patient's suffix and previous name MS tests are included in patient test group (5.2.09 and 6.2.09) for uscore-4 and uscore-5

There's some other stuff that needs to happen with that, so I'm going to make a separate ticket for it.

onc_certification_g10_test_kit.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'rubyzip', '~> 2.3.2'
   spec.add_runtime_dependency 'smart_app_launch_test_kit', '0.1.5'
   spec.add_runtime_dependency 'tls_test_kit', '0.1.1'
-  spec.add_runtime_dependency 'us_core_test_kit', '0.2.5'
+  # spec.add_runtime_dependency 'us_core_test_kit', '0.2.5'
```
yunwwang, line 21: Do we need this line or not?

Jammjammjamm: We'll be restoring it before we do a real release.

yunwwang (approved):

PR #212, opened 2022-08-15
FI-1678: Disable obsolete attestations

Jammjammjamm:
Attestations for `Patient.name.suffix` and `Patient.period` are only needed when using US Core 3 because the requirements are included in the profiles for US Core 4 & 5 and tested automatically.


yunwwang (approved):
Works as expected

PR #213, opened 2022-08-15
FI-1648: Add negative PKCE tests

Jammjammjamm:
This branch adds four scenarios to verify that servers are correctly verifying PKCE parameters.

### DIFF
--- a/lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
+++ b/lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
@@ -56,8 +56,8 @@ module ONCCertificationG10TestKit
       Launch](http://hl7.org/fhir/smart-app-launch/STU2/app-launch.html#launch-app-standalone-launch)
       Sequence, verifies that servers properly support PKCE.  It does this by ensuring the launch fails
       in the case where the client sends an invalid PKCE `code_verifier`.
-      
-      This group performs four launches with various forms of an invalid `code_verifier` 
+
+      This group performs four launches with various forms of an invalid `code_verifier`
       (e.g. incorrect `code_verifier`, blank `code_identifier`) and verifies that these do
       not result in a successful launch.  Testers can expect to be prompted four times
       that a redirect will occur in this test.
--- a/lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
+++ b/lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
@@ -1,0 +1,244 @@
+module ONCCertificationG10TestKit
+  class InvalidSMARTTokenRequestTest < Inferno::Test
+    title ' OAuth token exchange fails when supplied invalid code_verifier'
+    description %(
+      If the request failed verification or is invalid, the authorization
+      server returns an error response.
+    )
+    uses_request :redirect
+    id :invalid_pkce_request
+
+    input :code, :use_pkce, :pkce_code_verifier, :client_id, :client_secret, :smart_token_url
+
+    def modify_oauth_params(oauth_params)
+      oauth_params
+    end
+
+    run do
+      skip_if request.query_parameters['error'].present?, 'Error during authorization request'
+
+      oauth2_params = {
+        grant_type: 'authorization_code',
+        code: code,
+        redirect_uri: config.options[:redirect_uri]
+      }
+
+      oauth2_headers = { 'Content-Type' => 'application/x-www-form-urlencoded' }
+
+      if client_secret.present?
+        client_credentials = "#{client_id}:#{client_secret}"
+        oauth2_headers['Authorization'] = "Basic #{Base64.strict_encode64(client_credentials)}"
+      else
+        oauth2_params[:client_id] = client_id
+      end
+
+      modify_oauth_params(oauth2_params)
+
+      post(smart_token_url, body: oauth2_params, name: :token, headers: oauth2_headers)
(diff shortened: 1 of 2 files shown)


lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
```diff
@@ -0,0 +1,244 @@
...
+
+    # no code_verifier
+
+    test from: :smart_app_redirect_stu2,
+         id: :smart_no_code_verifier_redirect
+    test from: :smart_code_received,
+         id: :smart_no_code_verifier_code_received
+
+    test from: :invalid_pkce_request do
+      title ' OAuth token exchange fails when no code_verifier is given'
```
arscan, line 190:
```suggestion
      title 'OAuth token exchange fails when no code_verifier is given'
```

Extra space here and in a whole bunch others if you search for `' OAuth`

lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
```diff
@@ -0,0 +1,244 @@
...
+      id :smart_no_verifier_token_request
+    end
+
+    # code_verifier=
+
+    test from: :smart_app_redirect_stu2,
```
arscan, line 196: Can we customize the wait message provided to the user so they don't see the same redirect message each time.  I would think people would think it is in an error and it is stuck in a loop? So for each, maybe prepend a line like 'Invalid PKCE code_verifier 3/4: This launch will provide a blank code_verifier and verify the server does not issue a bearer token. \n\n <and then follow with normal launch message>'.

lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
```diff
@@ -0,0 +1,244 @@
...
+        oauth2_headers['Authorization'] = "Basic #{Base64.strict_encode64(client_credentials)}"
+      else
+        oauth2_params[:client_id] = client_id
+      end
+
+      modify_oauth_params(oauth2_params)
+
+      post(smart_token_url, body: oauth2_params, name: :token, headers: oauth2_headers)
+
+      assert_response_status(400)
```
arscan, line 39: Are we sure this has to be a 400?  Should we instead just ensure it doesn't succeed (e.g. a 20x and there is no response body with the bearer token in it?).  Is the language in the spec strong enough to dictate exactly how the system reports a problem back to the client?

Jammjammjamm: Hmm, I think 401 may also be acceptable for these particular cases.

lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
```diff
@@ -0,0 +1,305 @@
...
+      [Standalone
+      Launch](http://hl7.org/fhir/smart-app-launch/STU2/app-launch.html#launch-app-standalone-launch)
+      Sequence, does not work in the case where the client sends an invalid PKCE
+      `code_verifier`. This group performs four launches without a valid
+      `code_verifier` and verifies that these do not result in a successful
+      launch.
+
+      This test is not included as part of a regular SMART Launch Sequence
+      because some servers may not accept an authorization code after it has
+      been used unsuccessfully in this manner.
```
arscan, line 64:
```suggestion
      The #{title} Group verifies that a SMART Launch Sequence, specifically the
      [Standalone
      Launch](http://hl7.org/fhir/smart-app-launch/STU2/app-launch.html#launch-app-standalone-launch)
      Sequence, verifies that servers properly support PKCE.  It does this by ensuring the launch fails
      in the case where the client sends an invalid PKCE `code_verifier`.
      
      This group performs four launches with various forms of an invalid `code_verifier` 
      (e.g. incorrect `code_verifier`, blank `code_identifier`) and verifies that these do
      not result in a successful launch.  Testers can expect to be prompted four times
      that a redirect will occur in this test.

      This test is not included as part of the Single Patient App group
      because there is no way for a client to infer that PKCE is supported on the server
      properly without performing extra launches.  Attempting to verify this within the
      same launch cannot be done because some servers may not accept an authorization code
      after it has been used unsuccessfully in this manner.
```

lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
arscan: Ok, I feel like we always continually back off the specific status code, but 400,401 is good enough for now :)

lib/onc_certification_g10_test_kit/smart_invalid_pkce_group.rb
```diff
@@ -0,0 +1,310 @@
...
+    description %(
+      # Background
+
+      The #{title} Group verifies that a SMART Launch Sequence, specifically the
+      [Standalone
+      Launch](http://hl7.org/fhir/smart-app-launch/STU2/app-launch.html#launch-app-standalone-launch)
+      Sequence, verifies that servers properly support PKCE.  It does this by ensuring the launch fails
+      in the case where the client sends an invalid PKCE `code_verifier`.
+      
+      This group performs four launches with various forms of an invalid `code_verifier`
```
arscan, line 60:
```suggestion

      This group performs four launches with various forms of an invalid `code_verifier`
```

Fix rubocop.

arscan (approved):

PR #214, opened 2022-08-16
FI-1673 Initial pass at bulk data us core 5 validation.

arscan:
Adds additional resource for us_core_5:
* ServiceRequest

Adds additional optional resource types for us_core_5 (omits if none provided):
* RelatedPerson
* PractitionerRole
* QuestionairreResponse

Updates logic for Condition and Observation for choosing relevant profiles.  For Observation relating to SDOH, it only validates against the SDOH profile and not the parent 'survey' profile.

Updated all descriptions as well.

I didn't add self-tests.  I wasn't sure how to handle this with Features and options.  Maybe this is something we can circle back on next month?

To test, update .env to have:

```
US_CORE_4_ENABLED=true
```

And take a look at the US Core 5 vs. US Core 4/3 tests when enabled.

### DIFF
--- a/lib/onc_certification_g10_test_kit/profile_selector.rb
+++ b/lib/onc_certification_g10_test_kit/profile_selector.rb
@@ -1,5 +1,5 @@
 module ONCCertificationG10TestKit
-  module ProfileGuesser
+  module ProfileSelector
     def extract_profile(profile)
       case profile
       when 'Medication'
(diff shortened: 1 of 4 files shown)


lib/onc_certification_g10_test_kit/profile_selector.rb
```diff
@@ -57,6 +74,31 @@ def guess_profile(resource) # rubocop:disable Metrics/CyclomaticComplexity
...
+          return extract_profile('ObservationSocialHistory') if suite_options[:us_core_version] == 'us_core_5' &&
+                                                                resource_contains_category(resource, 'social-history',
+                                                                                           'http://terminology.hl7.org/CodeSystem/observation-category')
+
+          return extract_profile('ObservationSdohAssessment') if suite_options[:us_core_version] == 'us_core_5' &&
+                                                                 resource_contains_category(resource, 'sdoh',
+                                                                                            'http://terminology.hl7.org/CodeSystem/observation-category') && # rubocop:disable Layout/LineLength
+                                                                 resource_contains_category(resource, 'survey', 'http://terminology.hl7.org/CodeSystem/observation-category') # rubocop:disable Layout/LineLength
+
+          return extract_profile('ObservationSurvey') if suite_options[:us_core_version] == 'us_core_5' &&
```
yunwwang, line 96: I think we should combine ObservationSdohAssessment and ObservationSurvey. see my post in slack

arscan: All set

yunwwang (approved):

PR #215, opened 2022-08-17
FI-1676: Update resource access tests for US Core 5

Jammjammjamm:
This branch adds resource access tests for `ServiceRequest` when using US Core 5. US Core 5 also adds profiles for `QuestionnaireResponse` and `RelatedPerson`, but `QuestionnaireResponse` support is optional, and the only required search for `RelatedPerson` is by `id`, which we won't know immediately after a launch.

### DIFF
--- a/lib/onc_certification_g10_test_kit/unrestricted_resource_type_access_group.rb
+++ b/lib/onc_certification_g10_test_kit/unrestricted_resource_type_access_group.rb
@@ -27,6 +27,8 @@ module ONCCertificationG10TestKit
         * Practitioner
         * Organization
 
+      If testing against USCDI v2, ServiceRequest is also checked.
+
       For each of the resource types that can be mapped to USCDI data class or
       elements, this set of tests performs a minimum number of requests to
       determine that the resource type can be accessed given the scope granted.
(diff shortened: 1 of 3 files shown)


lib/onc_certification_g10_test_kit/unrestricted_resource_type_access_group.rb
```diff
@@ -371,5 +371,27 @@ def resource_group
...
+      test from: :g10_resource_access_test do
+        title 'Access to ServiceRequest resources granted'
+        description %(
+          This test ensures that access to the ServiceRequest is granted or
+          denied based on the selection by the tester prior to the execution of
+          the test. If the tester indicated that access will be granted to this
+          resource, this test verifies that a search by patient in this resource
+          does not result in an access denied result. If the tester indicated that
+          access will be denied for this resource, this verifies that search by
+          patient in the resource results in an access denied result.
```
arscan, line 385: We need to update all the descriptions in this set of tests because there is no "selection by the tester prior to the execution of the test" for the unrestricted type.  Seems like an oversight in v2.0.

arscan:
Description of the entire groups needs to be updated too, unfortunately.

arscan:
> Description of the entire group needs to be updated too

Not sure if you just haven't done this yet or if didn't realize this was something different than my inline comments.  But in case it was the latter, this needs to be updated to support USCDIv1 and USCDI v2.  I know its easier just to delete it, but this is an area where it is nice to have it written out whats happening and why we are checking certain resources:

https://github.com/onc-healthit/onc-certification-g10-test-kit/blob/7289c14852b44519b16957e3018aa357e393cec3/lib/onc_certification_g10_test_kit/restricted_resource_type_access_group.rb#L6-L50

Jammjammjamm:
Sorry, I missed that comment.

arscan (approved):

PR #222, opened 2022-08-18
FI-1689: SMART v2 scopes

Jammjammjamm:
This branch adds support for SMART v2 scopes. When using SMART v2, v2 scopes are required for for the Standalone, Limited, and EHR launches.

### DIFF
--- a/lib/onc_certification_g10_test_kit/limited_scope_grant_test.rb
+++ b/lib/onc_certification_g10_test_kit/limited_scope_grant_test.rb
@@ -39,9 +39,14 @@ module ONCCertificationG10TestKit
     end
 
     def scope_granting_access?(resource_type, scopes)
-      scopes.any? do |scope|
-        scope.start_with?("patient/#{resource_type}", 'patient/*') && scope.end_with?('*', 'read')
-      end
+      scopes
+        .select { |scope| scope.start_with?("patient/#{resource_type}", 'patient/*') }
+        .any? do |scope|
+          _type, resource_access = scope.split('/')
+          _resource, access_level = resource_access.split('.')
+
+          access_level.match?(/\A(\*|read|c?ru?d?s?\b)/)
+        end
     end
 
     run do
--- a/lib/onc_certification_g10_test_kit/smart_ehr_practitioner_app_group.rb
+++ b/lib/onc_certification_g10_test_kit/smart_ehr_practitioner_app_group.rb
@@ -142,7 +142,7 @@ module ONCCertificationG10TestKit
           ['openid', 'fhirUser', 'launch', 'offline_access']
         end
 
-        def scope_type
+        def required_scope_type
           'user'
         end
       end
@@ -261,7 +261,7 @@ module ONCCertificationG10TestKit
             ['openid', 'fhirUser', 'launch', 'offline_access']
           end
 
-          def scope_type
+          def required_scope_type
             'user'
           end
         end
--- a/lib/onc_certification_g10_test_kit/smart_ehr_practitioner_app_group.rb
+++ b/lib/onc_certification_g10_test_kit/smart_ehr_practitioner_app_group.rb
@@ -135,6 +135,9 @@ module ONCCertificationG10TestKit
           inputs: {
             requested_scopes: { name: :ehr_requested_scopes },
             received_scopes: { name: :ehr_received_scopes }
+          },
+          options: {
+            scope_version: :v1
           }
         )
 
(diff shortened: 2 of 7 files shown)


arscan:
Against our reference server this doesn't work.  It seems like the reference server is acting properly (even though we haven't updated it -- it just isn't sophisticated enough to really parse scopes to this depth it seems like).

Screen Shot 2022-08-19 at 10 15 18 AM

Jammjammjamm:
I missed that one. Should be fixed now.

lib/onc_certification_g10_test_kit/smart_ehr_practitioner_app_group.rb
```diff
@@ -135,14 +136,17 @@ class SmartEHRPractitionerAppGroup < Inferno::TestGroup
           inputs: {
             requested_scopes: { name: :ehr_requested_scopes },
             received_scopes: { name: :ehr_received_scopes }
+          },
```
radamson, line 139: EHR Practitioner App launch shows v1 scopes by default, even when SMARTv2 is selected

This looks like the culprit:
https://github.com/onc-healthit/onc-certification-g10-test-kit/blob/5b4bf5c9505391e9c26c4b1cf4c6fa2b02db34bc/lib/onc_certification_g10_test_kit/smart_ehr_practitioner_app_group.rb#L118-L127

Screen Shot 2022-08-19 at 4 14 00 PM

lib/onc_certification_g10_test_kit/smart_standalone_patient_app_group.rb
```diff
@@ -131,14 +131,17 @@ class SmartStandalonePatientAppGroup < Inferno::TestGroup
           inputs: {
             requested_scopes: { name: :standalone_requested_scopes },
             received_scopes: { name: :standalone_received_scopes }
+          },
+          options: {
+            scope_version: :v1
```
radamson, line 136: Applying the preset overwrites the scopes with v1 style scopes

lib/onc_certification_g10_test_kit/limited_scope_grant_test.rb
```diff
@@ -39,9 +39,14 @@ def possible_resources
...
-        scope.start_with?("patient/#{resource_type}", 'patient/*') && scope.end_with?('*', 'read')
-      end
+      scopes
+        .select { |scope| scope.start_with?("patient/#{resource_type}", 'patient/*') }
+        .any? do |scope|
+          _type, resource_access = scope.split('/')
+          _resource, access_level = resource_access.split('.')
+
+          access_level.match?(/\A(\*|read|c?ru?d?s?\b)/)
+        end
```
radamson, line 49: Our reference server doesn't allow us to limit scopes to specific resources when a `*` type scope is used. I think this is a change that might warrant it's own PR (or an update to the reference server to allow `*` scopes to be expanded and selected.

Screen Shot 2022-08-19 at 4 23 22 PM
Screen Shot 2022-08-19 at 4 21 11 PM
Screen Shot 2022-08-19 at 4 23 48 PM

radamson (commented):
Need to update the EHR Default and the presets override the scopes, but otherwise things worked as expected.

I ran through the launches:

Screen Shot 2022-08-19 at 4 28 48 PM

- Standalone & EHR - had unrelated required capability errors, but otherwise worked and the sent/received scopes looked right.
  - EHR I had to manually enter the `launch openid fhirUser offline_access user/*.rs` scope
- Limited Access - The reference server doesn't let me deny individual resources
- Public Client - ✅

lib/onc_certification_g10_test_kit/limited_scope_grant_test.rb
Jammjammjamm: I don't think that is a requirement.

lib/onc_certification_g10_test_kit/smart_standalone_patient_app_group.rb
Jammjammjamm: I don't think that is going to change before release.

lib/onc_certification_g10_test_kit/limited_scope_grant_test.rb
radamson: Ya not a requirement. It's really a reference server issue since we can't pass the limited scope test when '*' resource scopes are requested

lib/onc_certification_g10_test_kit/smart_ehr_practitioner_app_group.rb
Jammjammjamm: I updated this PR to make them consistent across all of the launches.

lib/onc_certification_g10_test_kit/limited_scope_grant_test.rb
Jammjammjamm: Yeah, I updated all of the defaults to not use wildcard scopes.

yunwwang (approved):

PR #223, opened 2022-08-19
Fi 1690 suite summary

arscan:
Adds suite summary.  Enable the feature flags, and a logical suite summary describing the SVAP options should be there.  I also removed references to v1.9 in the suite description, i think we are sufficiently past that now to remove references to it.


lib/onc_certification_g10_test_kit.rb
```diff
@@ -201,6 +196,20 @@ def self.well_known_route_handler
...
+      for Health Level 7 (HL7®) Fast Healthcare Interoperability Resources
+      (FHIR®) services seeking to meet the requirements of the Standardized API
+      for Patient and Population Services criterion § 170.315(g)(10) in the 2015
+      Edition Cures Update rule.
+
+      Systems may adopt later versions of standards than those named in the rule
+      as approved by the ONC Standards Version Advancement Process (SVAP).
+      Please select which approved version of each standard to use, and click
+      ‘Select Options’ to begin testing.
+
```
Jammjammjamm, line 210: Extra whitespace

Jammjammjamm (approved):

lib/onc_certification_g10_test_kit.rb
arscan, line 210:
```suggestion
      ‘Select Options’ to begin testing.
```

arscan: argh it won't let me just delete a line using a suggestion.

arscan: Fixed

PR #224, opened 2022-08-19
FI-1584 EHR Launch with patient scopes

radamson:
# Summary

Adds additional launch for patient-scoped EHR launches.

# Testing Guidance

There is a patient-scoped ehr launch group with the additional tests for smart v1 and v2.


lib/onc_certification_g10_test_kit/smart_ehr_patient_launch_group_stu2.rb
```diff
@@ -0,0 +1,94 @@
...
+      inputs: {
+        client_id: {
+          name: :ehr_patient_client_id
+        },
+        client_secret: {
+          name: :ehr_patient_client_secret
+        },
+        requested_scopes: {
+          name: :ehr_patient_requested_scopes,
+          default: 'launch openid fhirUser offline_access patient/Patient.rs',
```
Jack-Fraser16, line 34: So ... `Patient.rs` format in STU2 is the only difference? :sweat_smile:

Jack-Fraser16 (approved):

lib/onc_certification_g10_test_kit/smart_ehr_patient_launch_group_stu2.rb
Jammjammjamm: The only difference in scopes. The STU1 and 2 versions inherit from different base groups.

dhit-mdavis:
Hi @Jammjammjamm, can you walk thru the expected use case to have patient scopes under an EHR Practitioner Launch?

dhit-mdavis:
Thanks @Jammjammjamm. This makes sense logically. Another use case to consider.

dhit-mdavis:
Hi @Jammjammjamm , 1 final clarification here...is this essentially an EHR Patient Launch OR is this an EHR Practitioner Launch scoped to 1 patient?

Some clarity here would be extremely helpful. If it is an EHR Patient Launch, having the consistency in naming to match the EHR Practitioner Launch would be greatly appreciated.

Jammjammjamm:
> Patients apps that launch standalone
Provider apps that launch standalone

These scenarios are exactly the same (standalone launch).

> Patient apps that launch from a portal
Provider apps that launch from a portal

These scenarios are exactly the same (EHR launch).

That is why I said it doesn't matter. There are only two SMART workflows, standalone launch and EHR launch. Whether it is a "patient" app or a "provider" app has no impact on the SMART workflow.

dhit-mdavis:
Okay, so you're equating portal to EHR Launch.

Short answer is you can have either a patient app or a provider app from a portal (EHR Launch) perspective.
You did say this but I needed specific clarity here as I echo this to clients.

**The problem here is that** 
a Patient App launching from a portal = an EHR Launch from a patient portal
a Provider App launching from a portal = an EHR Launch from the EHR portal itself which only providers have access to

Jammjammjamm:
> an EHR Launch from a patient portal
an EHR Launch from the EHR portal itself which only providers have access to

These are exactly the same as far as SMART is concerned.

dhit-mdavis:
I think that's reasonable to say but it's not exactly accurate when it comes to implementing bc while SMART sees them as the same thing, there could be 2 different processes at the implementation level and will have 2 different configurations for 2 different apps

I do appreciate your perspective and info regardless though.

My main goal was to ensure that clients understand what's expected so I thank you for that.

PR #225, opened 2022-08-19
FI-1658: Add encounter context test to EHR launch group

Jammjammjamm:
Encounter is in USCDI v2, so the Encounter context should be tested when using US Core 5. If you run the EHR launch with US Core 5, you should see the encounter test.


radamson (commented):
`context-ehr-encounter` should be added to the ehr practitioner app group discover test when US Core v5 is selected.

Otherwise looks good.

yunwwang (approved):

yunwwang:
This fixed GitHub Issue #230

PR #242, opened 2022-08-31
FI-1700: Encode Param in _outputFormat Test

Jack-Fraser16:
Updated the export performer to take in a hash of params and encode each param before appending them to the url. 

To test, run the Multi-Patient API STU2 group and confirm that Test 8.2.08 passes as expected.

### DIFF
--- a/spec/onc_certification_g10_test_kit/bulk_data_group_export_stu2_spec.rb
+++ b/spec/onc_certification_g10_test_kit/bulk_data_group_export_stu2_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe ONCCertificationG10TestKit::BulkDataGroupExportSTU2 do
   describe '[Bulk Data Server supports "_outputFormat" query parameter] test' do
     let(:runnable) { group.tests.last }
     let(:long_format_req) do
-      stub_request(:get, "#{export_url}?_outputFormat=application/fhir+ndjson")
+      stub_request(:get, "#{export_url}?_outputFormat=application/fhir%2Bndjson")
         .to_return(status: 202, headers: { 'Content-Location' => polling_url })
     end
     let(:medium_format_req) do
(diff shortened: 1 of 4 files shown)


spec/onc_certification_g10_test_kit/bulk_data_group_export_stu2_spec.rb
```diff
@@ -34,7 +34,7 @@ def run(runnable, inputs = {})
   describe '[Bulk Data Server supports "_outputFormat" query parameter] test' do
     let(:runnable) { group.tests.last }
     let(:long_format_req) do
-      stub_request(:get, "#{export_url}?_outputFormat=application/fhir+ndjson")
+      stub_request(:get, "#{export_url}?_outputFormat=application/fhir%2Bndjson")
```
arscan, line 37: should `/` be URL encoded too?

spec/onc_certification_g10_test_kit/export_kick_off_performer_spec.rb
```diff
@@ -12,12 +12,12 @@ def http_clients
 RSpec.describe ONCCertificationG10TestKit::ExportKickOffPerformer do
   let(:token) { 'some_token' }
   let(:group_id) { 'some_group_id' }
-  let(:param) { '_outputFormat=ndjson' }
   let(:bulk_server_url) { 'https://www.example.com' }
   let(:polling_url) { 'https://www.some_polling_url.com' }
   let(:performer) { ExportKickOffPerformerTesterClass.new }
   let(:request) { Inferno::Entities::Request.new({ headers: [header] }) }
   let(:bulk_export_url) { "#{bulk_server_url}/Group/#{group_id}/$export" }
+  let(:params) { { _outputFormat: 'appplication/fhir+ndjson', _sort: 'sample+value' } }
```
arscan, line 20: Probably doesn't matter technically, but you misspelled `application` here.

spec/onc_certification_g10_test_kit/bulk_data_group_export_stu2_spec.rb
arscan: (same comment applies to all of these `stub_requests`)

Jack-Fraser16: Good catch. I'm assuming stubbing doesn't care if its encoded or not. Just changed and pushed up.

spec/onc_certification_g10_test_kit/export_kick_off_performer_spec.rb
Jack-Fraser16: :slightly_frowning_face:

arscan (approved):
Tried it out and it looks good!

Issue #252, opened 2022-09-06
Add invalid token refresh test (FI-1706)


Jammjammjamm:
Section `AUT-PAT-23` of the test procedure states:

> The health IT developer demonstrates the ability of the Health IT Module to return an error response when supplied an invalid refresh token as specified in the implementation specification adopted in § 170.215(a)(3).

This is not currently tested, so tests need to be added to cover this.

PR #256, opened 2022-09-07
FI-1706: Add invalid refresh token test

Jammjammjamm:
This branch adds tests to check the response when trying to perform a token refresh with an invalid refresh token (see #252).


arscan (approved):
This works fine, **but if you leave it as-is you'll have to update the matrix numbering more than this** because you've bumped existing numbers around by adding this in the middle instead of the end.

https://github.com/onc-healthit/onc-certification-g10-test-kit/blob/81c6304a83759e7f221a03dba06e64f0f85a6c0f/lib/onc_certification_g10_test_kit/onc_program_procedure.yml#L929-L931


I would recommend putting the failure case at the end, so you've just added `.06` instead of splitting the success case of `.04, .05` and having to update test numbering.  And any time we update this matrix, remember that there might be others downstream updating their numbers too, so its better to keep this as light a touch as possible I think.

PR #265, opened 2022-09-12
Update US Core test kit

Jammjammjamm:
This branch updates the US Core test kit. The US Core 5 test suite now contains an optional group, so we filter those out when we import the groups for the single patient api. If you run the g10 suite with US Core 5, you should not see a group for QuestionnaireResponse.


arscan:
This doesn't update numbering for the matrix?  Just checking.

arscan (approved):
Confirmed that this does not exist.

I think the matrix should be ok here, because our US Core test mapping all aligns to such a small part of the test procedure.  But brought it up just in case that was an oversight... probably better to update the matrix's yml file in this PR if its necessary.

yunwwang (approved):

PR #272, opened 2022-09-16
FI-1718: Handle imaging result profile in bulk data

Jammjammjamm:
Adds support for the Observation Imaging Result profile in bulk data when using US Core 5.

### DIFF
--- a/lib/onc_certification_g10_test_kit/profile_selector.rb
+++ b/lib/onc_certification_g10_test_kit/profile_selector.rb
@@ -148,6 +148,12 @@ module ONCCertificationG10TestKit
           return extract_profile('ObservationSocialHistory')
         end
 
+        if suite_options[:us_core_version] == 'us_core_5' &&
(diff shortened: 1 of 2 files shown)


lib/onc_certification_g10_test_kit/profile_selector.rb
```diff
@@ -148,6 +148,12 @@ def select_profile(resource) # rubocop:disable Metrics/CyclomaticComplexity
           return extract_profile('ObservationSocialHistory')
         end
 
+        if suite_options[:us_core_version] == 'us_core_5' &&
```
yunwwang, line 151: Can we combine all the suite_options[:us_core_version] == 'us_core_5' logic together?

Jammjammjamm: We should have that, and it would affect for more than just this PR, so I'll make a follow-up ticket for that.

yunwwang (approved):

PR #284, opened 2022-10-02
patient/ServiceRequest.rs added to scopes

sidharthramesh:
patient/ServiceRequest.rs needs to be present here to get access to the ServiceRequest resource. Access to ServiceRequest is validated in the tests ahead.


arscan:
Thanks for the PR @sidharthramesh!  Unfortunately this is a little more complicated to fix because ServiceRequest is only relevant to US Core v5, so we don't want to request it when in US Core v3.11 or v4.  We've got a ticket on our side to try to come up with a strategy here, as I do think defaults are important.

arscan:
Closing this because we implemented this fix in a different way in #395.  Thanks for the effort though!

arscan:
My apologies, we addressed this issue for 'presets' (e.g. when pointing to our reference server), but not for defaults, which is different.  Regardless, it is a fairly involved update to ensure that the default scope isn't populated for US Core v3.1.1 and v4.0.0, and the solution will have to look a bit different than this, so I'm going to leave this closed.

PR #288, opened 2022-10-03
FI-1736: New terminology process

Jammjammjamm:
This branch updates the terminology build to:
- Only build from a single version of UMLS
- Only build terminology for required bindings
- Report code system versions in the UI
Screen Shot 2022-10-03 at 1 47 36 PM

I made the changes to the code as minimal as possible, so there are places where we are now doing things like iterating over single element arrays.


terminology_compose.yml
```diff
@@ -4,6 +4,9 @@ services:
     build:
       context: .
       dockerfile: Dockerfile.terminology
+    # Uncomment the line below if using an arm mac. For some reason
+    # metamorphosys isn't working in arm docker images.
+    # platform: 'linux/amd64'
```
Jammjammjamm, line 9: This lets the build get farther, but unfortunately doesn't seem to actually make it work on an m1.

arscan (approved):

Issue #161, opened 2022-07-07
Bulk Inferno test failure:   "5.3.01 Bulk Data Server is secured by transport layer security"  for AWS S2 signed URLs (FI-1684)


MothyTim:
Hi,
We are having an issue with passing this 1 5.3.01 test case in the bulk data Inferno tool.

The implementation we have set up returns AWS S3 signed URLs for the resource links in the Bulk Data Status request response and used for the subsequent fetching of the resources.    The AWS S3 bucket has a policy to only permit TLS 1.2 version.    As demonstrated with curl it properly rejects 1.1 access:

curl '<signed AWS S3 url involved>' --tlsv1.1 --tls-max 1.1
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>R<request id involved></RequestId><HostId><host involved></HostId></Error>

Using curl with 1.2 properly downloads the resource data.

But the inferno tool fails the test:

error: | Server incorrectly allowed TLS 1.0 connection.
error: | Server incorrectly allowed TLS 1.1 connection.

Is inferno only checking the AWS  host and not the bucket and it's policy?

The HL7 clearly states the AWS signed URLs are compliant since our implementation sets requiredAccessToken to false.

From specs:   https://hl7.org/fhir/uv/bulkdata/STU1.0.1/export/index.html

requiredAccessToken:  indicates whether downloading the generated files requires a bearer access tokenValue SHALL be true if both the file server and the FHIR API server control access using OAuth 2.0 bearer tokens. Value MAY be false for file servers that use access-control schemes other than OAuth 2.0, such as downloads from Amazon S3 bucket URLs or verifiable file servers within an organization's firewall.

Is there an issue with Inferno certification tool not properly handling the AWS S3 scenario for 5.3.01?

All the other tests are passing and the data is properly returned in the subsequent tests using the signed URLs.

Regards, Tim

yunwwang:
@MothyTim 

Thank you for raise this issue to us.

Inferno test 5.3.01 uses the first output.url in the Status Complete Response of 5.2.05. 

Please verify if that url is TLS 1.2 only.

~ Yunwei

arscan:
It sounds like this is similar to #154 and https://github.com/onc-healthit/inferno-program/issues/304

[...]

PR #289, opened 2022-10-04
FI-1684: Update TLS tests

Jammjammjamm:
This branch updates the TLS tests to not fail servers which allow connections over prohibited versions (see #161). Servers which allow these connections are required to document how they prevent data from being sent over insecure connections in an attestation.

Screen Shot 2022-10-04 at 10 55 38 AM

- Run standalone and ehr launches, point one at inferno-dev, and the other at inferno-qa. The TLS tests should pass with warnings.
- Run the attestations. You should see the TLS error messages and that an explanation is required. The attestation will fail if no explanation is given.


yunwwang:
Tested in docker. Result is as expect. There is minor typo but I don't where the original text is. Please fix this

image

"thisconnection" (2nd line of screenshot) shall be "this connection"

yunwwang (approved):

PR #290, opened 2022-10-04
FI-1742: Static IDs

Jammjammjamm:
This branch creates a way to maintain static short ids. There is a yaml file which contains mappings from id -> short id. Unit tests will fail if any mappings are missing from the mapping file.


lib/onc_certification_g10_test_kit/short_id_map.yml
```diff
@@ -0,0 +1,1522 @@
...
+? g10_certification-g10_single_patient_api-us_core_v311_pediatric_bmi_for_age-us_core_v311_pediatric_bmi_for_age_patient_category_date_search_test
+: 4.17.02
+? g10_certification-g10_single_patient_api-us_core_v311_pediatric_bmi_for_age-us_core_v311_pediatric_bmi_for_age_patient_category_search_test
+: 4.17.03
+g10_certification-g10_single_patient_api-us_core_v311_pediatric_bmi_for_age-us_core_v311_pediatric_bmi_for_age_read_test: 4.17.04
+? g10_certification-g10_single_patient_api-us_core_v311_pediatric_bmi_for_age-us_core_v311_pediatric_bmi_for_age_provenance_revinclude_search_test
+: 4.17.05
+g10_certification-g10_single_patient_api-us_core_v311_pediatric_bmi_for_age-us_core_v311_pediatric_bmi_for_age_validation_test: 4.17.06
+g10_certification-g10_single_patient_api-us_core_v311_pediatric_bmi_for_age-us_core_v311_pediatric_bmi_for_age_must_support_test: 4.17.07
+? g10_certification-g10_single_patient_api-us_core_v311_pediatric_bmi_for_age-us_core_v311_pediatric_bmi_for_age_reference_resolution_test
```
arscan, line 345: Why the question mark in front of some of the lines?  I figured maybe it had to do with some tests be suppressed because they are optional, but that doesn't look like it?

Jammjammjamm: I'd never seen that either before this. Apparently it's a way in yaml to have a complex key. You could use it to have something like an array as a key. I think it switches to that format when the keys reach a certain length. I don't know the particulars about why that means they need the special format keys.

arscan (approved):

PR #294, opened 2022-10-07
Fix CodeSystem metadata path

Jammjammjamm:
If you look at the g10 config info messages on inferno-dev, you should see this, which means this works:

Screen Shot 2022-10-07 at 9 38 27 AM


yunwwang (approved):
Looks good on infenro-dev

PR #310, opened 2022-10-19
FI-1763 Update README with SVAP versions and more info.

arscan:
Updates README to include all approved versions of standards and provides link to walkthrough, reference server, and onc-hosted instance.


Jammjammjamm (approved):
Looks good. There are some long lines that could use wrapping.

Issue #320, opened 2022-10-28
Inferno doesnt validate terminology codes (FI-1780)


Nbhatin:
I am running g10 suite for uscore3.1 . I had invalid codes and display for Patient Race, Ethnicity, language , birthsex etc
But Inferno passed the patient tests.

Does Inferno validate the codes with system ? eg if incorrect snomed code is provided for a profile, should it fail ?
I tried giving incorrect snomed code for Patient smoking observation as well, but the test passed.

Patient profile -
{
    "resourceType": "Patient",
    "id": "999321",
    "meta": {
        "versionId": "1",
        "lastUpdated": "2022-10-28T12:55:24.919+00:00",
        "source": "#e9fbb090-9c4d-41",
        "profile": [
            "http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient"
        ]
    },
    "extension": [
        {
            "url": "http://hl7.org/fhir/us/core/StructureDefinition/us-core-race",
            "extension": [
                {
                    "url": "ombCategory",
                    "valueCoding": {
                        "system": "urn:oid:2.16.840.1.113883.6.238",
                        "code": "123",
                        "display": "Black"
                    }
                },
                {
                    "url": "text",
                    "valueString": "Black"
                }
            ]
        },
        {
            "url": "http://hl7.org/fhir/us/core/StructureDefinition/us-core-ethnicity",
            "extension": [
                {
                    "url": "ombCategory",
                    "valueCoding": {
                        "system": "urn:oid:2.16.840.1.113883.6.238",
                        "code": "ABCD",
                        "display": "ABCD"
                    }
                },
                {
                    "url": "text",
                    "valueString": "Not Hispanic or Latino"
                }
            ]
        },
        {
            "url": "http://hl7.org/fhir/StructureDefinition/patient-mothersMaidenName",
            "valueString": "Lorrie Murphy561"
        },
        {
            "url": "http://hl7.org/fhir/us/core/StructureDefinition/us-core-birthsex",
            "valueCode": "ZZZ"
        },
        {
            "url": "http://hl7.org/fhir/StructureDefinition/patient-birthPlace",
            "valueAddress": {
                "city": "Berlin",
                "state": "Berlin",
                "country": "XYZ"
            }
        },
        {
            "url": "http://synthetichealth.github.io/synthea/disability-adjusted-life-years",
            "valueDecimal": 24.285486783792628
        },
        {
[...]

PR #324, opened 2022-10-28
FI-1780: Fix terminology checking

Jammjammjamm:
This fixes terminology validation.

If you run the tests against inferno-dev, you should see three terminology errors in the patient tests that you don't see on `main` (see #320).


arscan (approved):

arscan:
Confirmed tests that pass on -qa fail on -dev.

Issue #327, opened 2022-10-31
Clinical Test Observations not correctly identified in Bulk Data tests


Jammjammjamm:
When using US Core 5, Clinical Test Observations are not identified. See https://chat.fhir.org/#narrow/stream/179309-inferno/topic/Multi-Patient.20Authorization.20and.20API.20STU2.20.288.2E3.2E16.29

agasthahyd06:
Is the fix pushed to Inferno

Jammjammjamm:
It should be available on inferno-dev.healthit.gov, but it won't be on the main site until the official release next week.

PR #328, opened 2022-10-31
FI-1781: Fix clinical test observations in bulk data

Jammjammjamm:
Clinical test observations in US Core 5 were not being correctly identified in the bulk data tests due to an incorrect system url (see #327).


arscan (approved):
Looks good.  I think I missed this because this happens to be a special case category and they created a whole new CodeSystem to accommodate it. I checked on the single patient api side, and I do not think that it should be a problem there because it just gets pulled automatically from the profile.

Issue #330, opened 2022-10-31
EHR Launch with Patient Scopes Tests do not validate that patient scopes are received (FI-1779)


Jammjammjamm:
The EHR Launch with Patient Scopes tests lock the requested scopes so that patient-level scopes must be requested, but do not verify that patient-level scopes are granted. We need to check that behavior and improve the description so that it is clear that an instance of Inferno registered with patient-level scopes needs to be used for these tests.

See: https://chat.fhir.org/#narrow/stream/179309-inferno/topic/9.2E9.2E08.20EHR.20Launch.20w.20Patient.20Scope.3A.20Warning.20a.20Failure.3F

PR #332, opened 2022-11-02
FI-1779: Verify EHR patient launch scopes

Jammjammjamm:
This branch adds tests to verify that patient-level scopes are received during the EHR Launch with Patient Scopes (see #330).


arscan (approved):
I think I had mentioned this before, but we used to allow systems to only support the wildcard syntax if they wanted.  This test group prevents that.  Not new to this PR, so this is fine, but I think we should keep an eye out for any complaints about this.

Otherwise looks good to me.

PR #335, opened 2022-11-07
FI-1787: Update to use core dev improvements (docker-compose v2 required, systems may need to upgrade)

Jammjammjamm:
https://github.com/inferno-framework/inferno-template/releases/tag/v0.0.3


arscan (approved):
Works fine for me now that i have docker-compose v2 installed.  I put a not in the title of the PR so we can remember to add a note in the next release that systems may need to update to docker-compose v2.

Issue #340, opened 2022-11-09
Fix Condition validation in US Core 5


Jammjammjamm:
Condition terminology validation is broken is US Core 5 due to the slicing by required binding on `Condition.category`.

See: https://chat.fhir.org/#narrow/stream/179309-inferno/topic/6.20Single.20Patient.20API.20.28US.20Core.205.2E0.2E1.29

PR #341, opened 2022-11-09
FI-1789: Add special case for us core 5 Condition.category terminology

Jammjammjamm:
`Condition/260` on  inferno-dev has been updated to include the US Core 5 categories. The `Condition Problems and Health Concerns` tests will fail against inferno-dev on `main`, but should pass on this branch (see #340).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    onc_certification_g10_test_kit (3.3.0)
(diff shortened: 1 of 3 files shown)


arscan (approved):
inferno-dev now does not provide that error, though it does provide others as expected.

I take it adding a 'unit' test for this doesn't make sense?

Should we put this in a prerelease and ask folks to try it out before an official release?

Issue #361, opened 2022-11-29
Generalize the fix for Condition validation in US Core 5 (FI-1792)


yunwwang:
This is a generalized solution to GitHub Issue #340 fixed in version 3.3.1. This fix applies to any profile with mandatory slice with required binding strength

PR #362, opened 2022-11-29
FI-1792 Terminology Validator handles binding with required binding slice

yunwwang:
This is part of fix for GitHub Issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/361.

This PR relates to US Core Test Kit PR: https://github.com/inferno-framework/us-core-test-kit/pull/94

This PR changes:
* Replace the validate logic for us-core-5-condition with a more general logic for any binding in a mandatory slice with `required` binding strength
* Add unit test

Testing:
Verify all unit tests pass

### DIFF
--- a/lib/onc_certification_g10_test_kit/terminology_binding_validator.rb
+++ b/lib/onc_certification_g10_test_kit/terminology_binding_validator.rb
@@ -37,11 +37,11 @@ module ONCCertificationG10TestKit
 
       return if valid_binding.present?
 
-      system = binding_definition[:system].presence || 'the declared CodeSystem'
+      system = binding_definition[:system].presence || 'the declared Value Set'
 
       error_message = %(
-        #{resource_type}/#{resource.id} at #{resource_type}.#{binding_definition[:path]} does not
-        contain a valid code from #{system} Value Set.
+        #{resource_type}/#{resource.id} at #{resource_type}.#{binding_definition[:path]}
+        does not contain a valid code from #{system}.
       )
 
       validation_messages << {
--- a/lib/onc_certification_g10_test_kit/terminology_binding_validator.rb
+++ b/lib/onc_certification_g10_test_kit/terminology_binding_validator.rb
@@ -18,11 +18,6 @@ module ONCCertificationG10TestKit
       @validation_messages = []
     end
 
-    def us_core_5_condition_category?
-      binding_definition[:path] == 'category' &&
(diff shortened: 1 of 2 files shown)


lib/onc_certification_g10_test_kit/terminology_binding_validator.rb
```diff
@@ -18,34 +18,30 @@ def initialize(resource, binding_definition)
...
+      return if valid_binding.present?
+
+      system = binding_definition[:system].presence || 'the declared CodeSystem'
 
       error_message = %(
-        #{resource_type}/#{resource.id} does not contain a `category` from the
-        `http://hl7.org/fhir/us/core/ValueSet/us-core-problem-or-health-concern`
-        ValueSet.
+        #{resource_type}/#{resource.id} at #{resource_type}.#{binding_definition[:path]} does not
+        contain a valid code from #{system} Value Set.
```
Jammjammjamm, line 44: This could end up being `contain a valid code from the declared CodeSystem Value Set` which doesn't seem quite right.

spec/onc_certification_g10_test_kit/terminology_binding_validator_spec.rb
```diff
@@ -212,5 +212,73 @@
...
+        expect(result).to be_an(Array)
+        expect(result.length).to eq(0)
+      end
+
+      it 'fails when required_bidning_slice is not specified' do
+        binding_definition.delete(:required_binding_slice)
+        result = described_class.validate(resource, binding_definition)
+
+        expect(result).to be_an(Array)
+        expect(result.length).to eq(1)
```
Jammjammjamm, line 280: Could you do an additional check here beyond just the length to demonstrate that an error is present?

spec/onc_certification_g10_test_kit/terminology_binding_validator_spec.rb
```diff
@@ -212,5 +212,73 @@
...
+      end
+
+      it 'passes resource with both required binding code and non required binding code' do
+        result = described_class.validate(resource, binding_definition)
+
+        expect(result).to be_an(Array)
+        expect(result.length).to eq(0)
+      end
+
+      it 'fails when required_bidning_slice is not specified' do
```
Jammjammjamm, line 275:
```suggestion
      it 'fails when required_binding_slice is not specified' do
```

Jammjammjamm (approved):

PR #368, opened 2022-12-12
FI-1670: Make patient id list optional for bulk data tests

Jammjammjamm:
This branch fixes the patient id list input for bulk data so that it is now optional as stated in its description.


arscan (approved):
I ran the tests with the field blank to ensure the test can handle it, and it properly ommitted the test.

Issue #369, opened 2022-12-12
Mime-type codes with charsets are not being validated correctly


Jammjammjamm:
See https://chat.fhir.org/#narrow/stream/179309-inferno/topic/MIME.20type.20validation.20failing.20when.20charset.20parameter.20included

Codes like `text/plain; charset=UTF-8` should be accepted but are failing validation.

PR #370, opened 2022-12-12
FI-1826: Handle validation of mime-type codes without a system

Jammjammjamm:
Fix a bug prevented mime-types with charsets from failing terminology validation when they don't have a system (see #369).


lib/inferno/terminology/terminology_validation.rb
```diff
@@ -6,7 +6,8 @@ module Terminology
     module TerminologyValidation
       # Code systems to "preprocess" prior to validation, and the function to use
```
arscan, line 7:
```suggestion
      # CodeSystems/Valuesets to "preprocess" prior to validation, and the function to use
```

arscan (approved):
Might be good to update the comment to indicate that preprocess methods apply to both VS and CS, as I put in the suggestion.

Issue #382, opened 2022-12-20
Add attestation to ensure that systems are capable of extending access beyond the minimum 3 months when requested


arscan:
The intent of the criterion is to ensure that systems can extend the length of authorized access for clients that receive refresh tokens, for an a minimum of 3 additional months, when requested on subsequent connections.  Users can always revoke these tokens if they would like to stop providing access to the client.

See: https://www.ecfr.gov/current/title-45/subtitle-A/subchapter-D/part-170/subpart-C#p-170.315(g)(10)(v)(A)(2)(ii)

On "Subsequent connections" for  "Authentication and authorization for patient and user scopes", "A Health IT Module's authorization server must issue a refresh token valid for a new period of no less than three months to applications capable of storing a client secret. ".

So for subsequent authentication/authorization connections (i.e. using the refresh token to receive a new access token), there must be a "new period" of at least 3 months that the a new (or the existing) refresh token is valid for.

This is also represented in the test procedure under "subsequent connections": https://www.healthit.gov/test-method/standardized-api-patient-and-population-services#test_procedure

> Subsequent Connections: Authentication and Authorization for Patient and User Scopes AUT-PAT-22: The health IT developer demonstrates the ability of the Health IT Module to issue a refresh token valid for a new period of no shorter than three months without requiring re-authentication and re-authorization when a valid refresh token is supplied by the application according to the implementation specification adopted in § 170.215(a)(3)

Note that this only applies to applications capable of storing a client secret, so the attestation will include that caveat.

The current 'Visual Inspection' section does not have an explicit question regarding this -- it only has a question regarding the duration of the refresh token, and not if that duration can be be extended by issuing a new refresh token or extending the life of the existing refresh token: "9.10.05 Health IT Module attested that it is capable of issuing refresh tokens that are valid for a period of no shorter than three months."

Add a new Visual Inspection check specific for this based on the language in the test procedure.

PR #386, opened 2023-01-05
fi-1748 Adds attestation to ensure that systems are capable of extending authorized access of refresh tokens (Minor-version change)

arscan:
Adds attestation regarding extending the life of refresh tokens. I based the language of the attestation from the test procedure, and provide more details in the 'About' section.

Screen Shot 2023-01-04 at 9 22 09 PM

Language from the test procedure:

> AUT-PAT-22: The health IT developer demonstrates the ability of the Health IT Module to issue a refresh token valid for a new period of no shorter than three months without requiring re-authentication and re-authorization when a valid refresh token is supplied by the application according to the implementation specification adopted in § 170.215(a)(3).

I felt like it would be useful to mention that this is different than the refresh token duration attestation in the  'about' section, as I figured that would be a common question.  That one is:

> 9.10.05 Health IT Module attested that it is capable of issuing refresh tokens that are valid for a period of no shorter than three months. 

Relevant github ticket: https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/382

### DIFF
--- a/lib/onc_certification_g10_test_kit/onc_program_procedure.yml
+++ b/lib/onc_certification_g10_test_kit/onc_program_procedure.yml
@@ -960,7 +960,7 @@ procedure:
           implementation specification adopted in § 170.215(a)(3).
         inferno_supported: 'yes'
         inferno_tests:
-          - 9.10.15
+          - 9.10.16
         inferno_notes: |
           Inferno cannot verify the three month token expiration requirement
           automatically during the token refresh tests, but the tester can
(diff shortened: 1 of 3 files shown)


Jammjammjamm (approved):

yunwwang:
Is it necessary to include the last sentence about "previous attestation"?

arscan:
> Is it necessary to include the last sentence about "previous attestation"?

I'm not sure.  I added it because I personally feel like it is confusing to have 2 attestations about refresh token duration, so I tried to provide a bit more information.  I updated it to be "a separate attestation" to avoid any ordering issues, which makes a bit more sense.

Note that this iis only shown if you click into the test to see the About tab.

Issue #385, opened 2023-01-04
Check bulk export status of cancelled job in bulk v2


Jammjammjamm:
> Following the delete request, when subsequent requests are made to the polling location, the server SHALL return a 404 Not Found error and an associated FHIR OperationOutcome in JSON format.

http://hl7.org/fhir/uv/bulkdata/STU2/export.html#bulk-data-delete-request

A test needs to be added for these requirements.

PR #387, opened 2023-01-05
FI-1765: Check status of cancelled bulk export (Minor Version Change)

Jammjammjamm:
Bulk data v2 requires specific behavior for status requests for cancelled bulk exports (see #385).

### DIFF
--- a/lib/onc_certification_g10_test_kit/bulk_data_group_export_stu1.rb
+++ b/lib/onc_certification_g10_test_kit/bulk_data_group_export_stu1.rb
@@ -286,9 +286,13 @@ module ONCCertificationG10TestKit
 
       include ExportKickOffPerformer
 
+      output :cancelled_polling_url
+
       run do
         perform_export_kick_off_request
         assert_response_status(202)
+
+        output cancelled_polling_url: request.response_header('content-location')&.value
 
         delete_export_kick_off_request
       end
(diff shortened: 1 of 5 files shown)


arscan:
This worked as expected on both the existing release of the reference server and the updated version to incorporate this requirement.  The language looks good and it is working as intended

However, I'm not seeing an update to the `onc_program_procedure.yml`.  Is this a gap in the test procedure?  I'm surprised that the TP missed this given how explicit the requirement is, but perhaps it can be justified under another existing step?

Jammjammjamm:
It could potentially go under the cancelling step or the status polling step.

arscan:
Based on discussions with ONC, we can map this to the existing Test Procedure step “DAT-PAT-12: The health IT developer demonstrates the ability of the Health IT Module to support a bulk data delete request according to the implementation specification adopted in § 170.215(a)(4).” Therefore, a test can be added to the ONC Certification (g)(10) Standardized API Test Kit for this requirement without altering the test procedure.

This requirement originates in the [Bulk Data Delete Request](http://hl7.org/fhir/uv/bulkdata/export.html#bulk-data-delete-request) section of the spec, even though it involves a request to the status polling endpoint.  So this request to the status polling endpoint can be considered a part of verifying the bulk data delete request works properly.

@Jammjammjamm Can you update the test procedure mapping in this PR?

arscan (approved):

Issue #389, opened 2023-01-20
Fix SMART IG Links in EHR Patient Launch


Jammjammjamm:
https://chat.fhir.org/#narrow/stream/179309-inferno/topic/EHR.20Launch.20with.20Patient.20Scopes.3A.20Wrong.20SMART.20Version.20Test.3F

The links in the EHR Patient Launch Group don't all point at the correct IG version. The group has links for STU 2, the patient scope test has links for STU 1.

PR #391, opened 2023-01-20
FI-1869: Fix smart IG links in EHR launch with patient scopes

Jammjammjamm:
Some links in the EHR Launch with Patient Scopes Group went to the incorrect IG version (see #389).

### DIFF
--- a/lib/onc_certification_g10_test_kit/patient_scope_test.rb
+++ b/lib/onc_certification_g10_test_kit/patient_scope_test.rb
@@ -3,11 +3,15 @@ module ONCCertificationG10TestKit
     title 'Patient-level scopes were granted'
     description %(
       Systems are required to support the `permission-patient` capability as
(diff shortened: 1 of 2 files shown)


arscan (approved):
This looks good 👍 

Should we have a separate ticket to align all of our naming so that version-specific things have version-specific filenames & modulenames to help avoid this kind of thing?

On the `smart_ehr_patient_launch_group.rb` it seems like we are stu1 specific but that isn't indicated in the file name or the module name.  I just bring it up because i had to confirm that there was a _stu2.rb file in that directory to verify that there is a stu2 group too.

Jammjammjamm:
STU 1 things generally don't specify their version in their filenames or ids because that would have been a breaking change. The EHR Patient Launch came later so it could have included STU 1, but then it wouldn't have been consistent with the other smart tests.

PR #395, opened 2023-01-25
FI-1878: Multiple version preset

Jammjammjamm:
This branch updates the reference server preset to use the new functionality which allows changing values based on the selected options.


arscan (approved):

arscan:
This fixes the same problem that https://github.com/onc-healthit/onc-certification-g10-test-kit/pull/284 PR was targeting, so we can close that after this is merged.

Jammjammjamm:
@arscan This does not duplicate #284 because that was targeting the default not the preset.

PR #417, opened 2023-03-15
FI-1921: Add reference server

Jammjammjamm:
This branch adds functionality to run the reference server locally.


arscan:
I think we need to clean up the language a little bit.  In reading this assuming no knowledge, I wouldn't know what a 'reference server' is in this context.  I can take a pass at updating the readme and will just push to this branch.

Issue #416, opened 2023-03-13
Relax MustSupport test for Patient's previous name and previous address (FI-1938)


yunwwang:
According to US Core JIRA ticket FHIR-40299 and the latest CCG update:

The HL7® Cross-Group Projects workgroup, through the [US Core 'Patch' Process](https://confluence.hl7.org/display/CGP/US+Core+%27Patch%27+Process) ticket [FHIR-40299](https://jira.hl7.org/browse/FHIR-40299), approved patching the US Core Patient Profile in US Core 3.1.1, US Core 4.0.0, and US Core 5.0.1. The USCDI data element “Patient Demographics: Previous Name” must be supported by including the capability to set the US Core Patient Profile element “Patient.name.use” to “old” or provide an end date in “Patient.name.period” element or support both. Additionally, the USCDI data element “Patient Demographics: Previous Address” must be supported by including the capability to set the US Core Patient Profile “Patient.address.use” element to “old” or provide an end date in “Patient.address.period” element or support both. Also, support for the US Core Patient Profile “Patient.address.period” element is not required for purposes of testing and certification.

Inferno will update Patient's MustSupport test accordingly

As a result, 
* For US Core v3.1.1
  * Attestation 9.10.12 will be removed
  * `Patient.address.period` will be removed from Patient's MustSupport test 4.2.09
  * These ONC USCDI requirement will be added to Patient's MustSupport test 4.2.09
    * `Patient.address.use` or `Patient.address.period.end` 
    * `Patient.name.use` or `Patient.name.period.end`
* For US Core v4.0.0 and v5.0.0, 
  * `Patient.address.period` will be removed from Patient's MustSupport test 5.2.09 and 6.2.09
  * These ONC USCDI requirement will be removed from Patient's MustSupport test 5.2.09 and 6.2.09
    * `Patient.address.use`
    * `Patient.name.period.end`
    * `Patient.name.use`
  * These ONC USCDI requirement will be added to Patient's MustSupport test 5.2.09 and 6.2.09
    * `Patient.address.use` or `Patient.address.period.end` 
    * `Patient.name.use` or `Patient.name.period.end`

We marked this issue "add constraint" because
* New test for ONC USCDI requirement is added for US Core v3.1.1
* If server uses Patient.address.period for previous address, Inferno used to test MustSupport at `Patient.address.period` but will change to `Patient.address.period.end`.

PR #418, opened 2023-03-17
FI-1941: Remove previous name attestation

Jammjammjamm:
See #416.


arscan (approved):
Looks good.  We need to make sure that this is released in concert with the improvements to the automated tests.  And since it may have downstream effects on supporting documentation from test labs, etc, that we call this out very specifically in the release notes.

Jammjammjamm:
> We need to make sure that this is released in concert with the improvements to the automated tests.

That's why this PR bumps the US Core Test Kit version.

PR #421, opened 2023-04-05
Release 3.7.0

Jammjammjamm:


arscan (approved):
Do you need to update the mapping file because we removed an attestation?  Also, we need to regenerate the matrix xls.  Otherwise seems good.

arscan:
Also, in looking at the release notes, everything is in 'future tense'.  Should it be in past tense now that we have actually done it?

Jammjammjamm:
We use present tense for release notes, but yes, good catch.

Issue #434, opened 2023-05-16
Failure message on 3.3.10  not accurate


arscan:
If a system grants scopes that are of the wrong "level" (patient/user), the error messages says that the issue is on 'received' scopes, when it reaslly is on 'granted' scopes.

For example, this was an error for scopes that were granted, received:
```
Requested scope patient/Patient.read does not follow the format: user/[ <ResourceType> | * ].[ read | * ]
```

It would be good to fix this for clarity.

PR #435, opened 2023-05-17
FI-1982: Fix bad scope message

Jammjammjamm:
This branch fixes a bug which cause a failed assertion to state that a requested scope was invalid when it's actually a received scope which is invalid.


arscan:
Addresses #434

PR #451, opened 2023-08-07
FI-2031: Integrate SMART App Launch Test Kit update

Jammjammjamm:
This branch updates the g10 suite to accomodate the new inputs included in the SMART App Launch Test Kit 0.3.0 release.

- For SMART App Launch 1 & 2:
  - A client secret was not required for the EHR Launch with Patient Scopes. This seems like an oversight, and it is now required to match all of the other groups (except public launch)
- For SMART App Launch 2 only:
  - Locked `Client Authentication Type` and `Client Authentication Encryption Method` inputs will be present on most SMART launch groups. The Invalid Token Request and Invalid PKCE Code Verifier groups are the exception and their inputs should appear unchanged.

Systems using the JSON API to run the g10 tests will need to update their API calls to include the new Client Auth Type & Encryption Method inputs if they using SMART App Launch 2. Presets should not need to be updated. The new inputs which need to be added are:
```json
    {
      "name": "client_auth_type",
      "value": "confidential_symmetric",
    },
    {
      "name": "client_auth_encryption_method",
      "value": "ES384",
    },
    {
      "name": "public_client_auth_type",
      "value": "public",
    },


arscan:
That locked input for the "Client Authentication Encryption Method" in smart v2 technically should be 'N/A', right?  (I know we can't do it, but in a perfect world it wouldn't be shown at all or have n/a. I just wonder if people might read into that incorrectly, like they are missing something by not using ES384 for symmetric authentication.

Could you update the description for that input in that specific case to show below such that it says "Only applicable 
for Confidential Asymmetric client authentication"

Screenshot 2023-08-07 at 4 27 36 PM

arscan:
Question 2: Is there a reason why Authorization Method is not locked in the limited access test in smart v2?  The description implies that it should be locked.  "All other inputs are locked to ensure the same app configuration as in the Standalone Patient App - Full Access test."

arscan (approved):

PR #455, opened 2023-08-14
Add list of applicable options to 'Inferno Tests' sheet in test procedure matrix

arscan:
There was some confusion about the content in the 'Inferno Tests' sheet because we don't indicate that some tests only apply to certain standards.  This PR adds a column to that sheet with that information.

I originally just created it to send to drummond to help explain how that sheet works, but it seems to make sense to include it going forward.


lib/onc_certification_g10_test_kit/tasks/generate_matrix.rb
```diff
@@ -188,6 +188,18 @@ def generate_test_procedure_worksheet # rubocop:disable Metrics/CyclomaticComple
         end
       end
 
+      # returns an array of options that apply to this test or group
+      def applicable_options(runnable)
+        runnable_and_parents = [runnable].tap do |parents|
+          while runnable.respond_to?(:parent) &&
+                runnable.respond_to?(:suite_option_requirements) &&
+                (runnable = runnable.parent)
```
Jammjammjamm, line 196:
```suggestion
                (runnable == runnable.parent)
```
I'm assuming you're intending to do a comparison rather than an assignment here

arscan: Nope, this is just bad code that i wrote quickly.  This is what allowed the while loop to traverse upwards.  I updated the code to be a bit simpler.

I'm not sure if this has to be this complicated though, I just wasn't sure how much options got pushed down to children in the test tree... so i just had it gather all relevant options for itself and all of its ancestors and then eventually do a 'uniq' to throw out any duplicates.

Jammjammjamm (approved):

PR #457, opened 2023-09-12
Release 3.8.1

Jammjammjamm:


arscan (approved):
When running tests on US COre 6.1.0, I"m getting a whole bunch of 'cannot resolve profile' for the 6.1.0 version.  I'm not seeing a 6.1.0 version of the test kit in the igs directory, I assume that's whats missing?  Approving becuase I assume all that needs doing is having that file in there.

Jammjammjamm:
Yeah, I don't think we should actually add that here, though, since we don't load that IG into the g10 validator.

nathanloyer:
@Jammjammjamm was there an issue with inferno core 0.4.19? I was wondering why it's locked on 0.4.18 now after these changes?

Jammjammjamm:
Yes, we inadvertently included a breaking change in 0.4.19. We released 0.4.20 yesterday to revert that change (#393 reverts the breaking change). None of the other changes in 0.4.19 seemed urgent enough to delay the g10 release, so we locked it to 0.4.18.

PR #458, opened 2023-09-29
select jwk by kid

nathanloyer:
I found that it was useful for my deployment to have multiple JWKS configured for the same algorithm. So I thought to add an optional input parameter for the bulk export tests to allow us to select which one to use via `kid`.

Let me know if you think this is/isn't generally useful for the community. We can continue to use our fork of the project if you don't want to merge this, but I'd prefer to get it merged. In my local testing I found this to behave identically if you do not set the kid input. If you do set it, then it is picking the correct one.

Let me know if there are automated test cases that should be updated or if you have any other suggestions.

Thanks

### DIFF
--- a/lib/onc_certification_g10_test_kit/authorization_request_builder.rb
+++ b/lib/onc_certification_g10_test_kit/authorization_request_builder.rb
@@ -12,7 +12,7 @@ module ONCCertificationG10TestKit
     end
 
     attr_reader :encryption_method, :scope, :iss, :sub, :aud, :content_type, :grant_type, :client_assertion_type, :exp,
-                :jti
+                :jti, :kid
 
     def initialize(
       encryption_method:,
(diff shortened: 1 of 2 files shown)


nathanloyer:
Hello @Jammjammjamm,

By all means there's no rush on this, but at some point could you or someone else review this and consider whether this is a feature worth merging? We can always continue to build our server from a fork if this doesn't seem generally useful to everyone.

For more context on why we did this, I wanted to test out different methods and encryption settings for generating JWKS for various test clients. If we can't pick which kid to use then we would have to have separate inferno deployments to test each.

Jammjammjamm:
We are currently focused on the SVAP update, but this is on our radar.

Jammjammjamm:
@nathanloyer I wanted to provide you with an update since this has been open for a while. Currently we only have bulk data tests here inside of the g10 test kit, but we are in the process of extracting those into their own test kit, similar to how we have standalone US Core and SMART App Launch tests. We are planning to incorporate the functionality in this PR as part of that work.

nathanloyer:
Thanks @Jammjammjamm. Is there a repo for that new test kit yet?

nathanloyer:
I see the kit support was merged in to the smart app launch test kit. I'll give that a try soon. Thanks

Jammjammjamm:
@nathanloyer We've just released a standalone bulk data test kit that incorporates this functionality: https://github.com/inferno-framework/bulk-data-test-kit

I'm not sure what the timeline will be for integrating this back into the g10 tests.

nathanloyer:
this has been resolved

Issue #461, opened 2023-10-11
Change CapabilityTest 1.05 to verify required profiles instead of resource types (FI-2163)


yunwwang:
US Core v4, v5, and v6 has a conformance requirement that:

* SHALL specify the full capability details from the US Core CapabilityStatement it claims to implement.
  * Declare support for the US Core Profile by including its official URL in the server’s CapabilityStatement.rest.resource.supportedProfile element

https://hl7.org/fhir/us/core/STU4/conformance-expectations.html#profile-support--interaction-support

Inferno CapabilityStatement test 1.05 only verifies the resource type, not supported profiles.

This is a breaking change for (g)(10) certification test when using US core v4.0.0 and v5.0.1

PR #462, opened 2023-10-11
FI-2163 Change required_resources to required_profiles

yunwwang:
This is part of solution to change the CapabilityStatement Test 1.05 to verify that server declared all supported profiles as specified in US core IG. Fix GitHub Issue #461 

Change Log:
Update single_patient test group to list all the required profiles


Jammjammjamm:
This seems good. Once we have a new US Core Test Kit release, we can bump the version in this PR and it should be good to go.

Jammjammjamm (approved):

PR #476, opened 2023-12-13
FI-2372: Fix preset

Jammjammjamm:
It doesn't make sense for the preset within the g10 test kit to point at the same host as Inferno. It should just point to our deployed reference server.


arscan (approved):
Ah, yeah, i noticed that when running the g10 tests against the new reference server locally, but forgot to try to track it down.  Thanks!

PR #478, opened 2023-12-20
FI-2334: Remove US Core 5.0.1

Jammjammjamm:
This branch removes US Core 5.0.1 as an option for certification testing. The tests themselves remain so that users can still see the results of previous test runs, but a message has been added to the US Core 5 Single Patient API group indicating that systems can no longer be certified using this version. The run button for the US Core 5 Single Patient API group has been removed, although the tests will still be run if a user runs the entire test suite at once.

To test:
- On `main`, start a g10 session using US Core 5, and run the Single Patient API tests
- Switch to this branch
- Reload that session and you see the results, the new message at the top of the Single Patient API tests, and that the run button no longer appears in that group
- If you start a new g10 session, you will see that US Core 5.0.1 is no longer an option
- All tests should still pass using other US Core versions. I checked all three, so feel free to just spot check.


arscan (approved):
The only minor issue i noticed was the fact that old us core 5 sessions don't mention what version of us core is being used here.  I guess we could generate this once and store it in the session in the future if we really cared, but i don't think it matters.

Screenshot 2023-12-27 at 10 44 17 PM

We need to make sure we remember to update the matrix, etc.

PR #488, opened 2024-02-06
FI-2460: Support for HL7 validator wrapper

dehall:
This PR enables support for the the g10-test-kit to use the original Inferno validator wrapper or the HL7 validator wrapper, toggled by the env var `USE_HL7_RESOURCE_VALIDATOR`.  From the user perspective, there should be no visible change. If I've done this right then it should be plug & play.

Summary of changes:
- Add HL7 validator service to docker-compose and nginx, commented out by default
- Change `validator` block in test suite to a conditional that selects `validator` or `fhir_resource_validator` based on a new env var `USE_HL7_RESOURCE_VALIDATOR`. Add `igs` setting to this block if using the HL7 validator.
- Add a `cli_context` block to specify validator settings:
   - `txServer nil` disables use of a TX server. Aligns with `DISABLE_TX` env var in the inferno validator
   - `displayWarnings true` sets any "display issue" messages to be warnings instead of errors.  Aligns with `DISPLAY_ISSUES_ARE_WARNINGS` env var in the inferno validator
   - `disableDefaultResourceFetcher true` disables fetching and validating against any unknown profiles found in a resource. (This is now a default in inferno-core, not yet released though. No harm in keeping this setting here temporarily or forever)
- Change version endpoint and expected version. (HL7 validator wrapper reports version as the the https://github.com/hapifhir/org.hl7.fhir.core release number)
- Ignore "URL does not resolve" errors - these were ignored natively in the Inferno validator but the mechanism to do that isn't exposed in the HL7 wrapper
- Ignore "[No server available]" errors - these were introduced somewhere in the core validator between versions 6.2.8 and 6.3.0 and occur when txServer is disabled. See screenshot below. 
- Update spec tests; see notes on #497 

New error messages that need to be ignored -- there are two messages that occur per usage of a mime type (CapabilityStatement, DocumentReference, some uses of DiagnosticReport)
image (12)


lib/onc_certification_g10_test_kit.rb
```diff
@@ -78,20 +79,29 @@ class G10CertificationSuite < Inferno::TestSuite
...
+        cli_context do
+          txServer nil
+          displayWarnings true
+          disableDefaultResourceFetcher true
+        end
+
+        case (us_core_version_requirement[:us_core_version])
+        when G10Options::US_CORE_3
+          igs('hl7.fhir.us.core#3.1.1')
+          us_core_message_filters = USCoreTestKit::USCoreV311::USCoreTestSuite::VALIDATION_MESSAGE_FILTERS
```
Jammjammjamm, line 94: I don't understand why this assignment was changed.

Jammjammjamm (commented):
This seems to be working well for me, but I expect your FHIR core PR will need to get in before we could use this in production. We should see what @arscan thinks as well. I think the next steps are something like the following:

- Make a subclass of the g10 suite, change its id, and have one use the current validator and one use the new validator so that we'll be able to support both during a transition period. I'm not sure whether we'll want the "real" g10 suite to use the new validator to ensure it gets broad testing or the current one to minimize any potential disruption. I guess this would need two version of the configuration checker as well.
- Update the US Core and SMART App Launch test suites to use the new validator so that we'll be ready to completely cut over and remove the current validator from this repo.

lib/onc_certification_g10_test_kit.rb
dehall: It just felt strange to have the `igs` line in the middle of the `case` when the result of that case is assigning to a separate variable. I'm probably going to revert this anyway because I suspect there's a better way to define the IG for each option.

Jammjammjamm: Having the `igs` lines there wouldn't interfere with the original assignment style.

dehall:
How will the "two suites" approach look from a user perspective? Will we show the two suites on the start page and the user has to pick one? 
In the meantime I'll press forward with cleanup here and then adding hl7 validator support to the other two test kits.

Jammjammjamm:
> How will the "two suites" approach look from a user perspective? Will we show the two suites on the start page and the user has to pick one?

We'll want to discuss that with @arscan

Jammjammjamm (approved):

PR #499, opened 2024-03-07
FI-2564 Update attestation for token revocation to include time limit.

arscan:
ONC has updated their requirement for token revocation to be clear that they must be complete within one hour of the patient request.  See [rule](https://www.federalregister.gov/d/2023-28857/p-3013).  This goes into effect March 11.

**NOTE: this changes the name of a test. I do not believe this has any significant downstream effects (e.g. breaking the test mapping across versions, for example).  But if it does let me know.**

Screenshot 2024-03-07 at 2 11 15 PM

### DIFF
--- a/config/presets/g10_reference_server_preset.json
+++ b/config/presets/g10_reference_server_preset.json
@@ -526,7 +526,7 @@
     {
       "name": "token_revocation_attestation",
       "type": "radio",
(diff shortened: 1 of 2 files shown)


Jammjammjamm (approved):

arscan:
Addresses #500

Issue #524, opened 2024-06-27
Inferno Test Cases 5.1.02 , 5.9.07, 5.10.08, 7.3.11 and 7.3.12 are failing with  Error: application/fhir+json or application/pdf mimetype not found for  Capability Statement for the FHIR server


Saiyed-GWHI:
**Description**: We are getting following errors  on https://inferno.healthit.gov/
**Test Case 5.1.02**
CapabilityStatement: CapabilityStatement.format[0]: The System URI could not be determined for the code 'application/fhir+json' in the ValueSet '[http://hl7.org/fhir/ValueSet/mimetypes|4.0.1](http://hl7.org/fhir/ValueSet/mimetypes%7C4.0.1)': include #0 has system urn:ietf:bcp:13 which could not be found, and the server returned error [Error from http://tx.fhir.org/r4: The code System "urn:ietf:bcp:13" has a grammar, and cannot be enumerated directly]
CapabilityStatement: CapabilityStatement.format[0]: The value provided ('application/fhir+json') was not found in the value set 'MimeType' ([http://hl7.org/fhir/ValueSet/mimetypes|4.0.1](http://hl7.org/fhir/ValueSet/mimetypes%7C4.0.1)), and a code is required from this value set (error message = The System URI could not be determined for the code 'application/fhir+json' in the ValueSet '[http://hl7.org/fhir/ValueSet/mimetypes|4.0.1](http://hl7.org/fhir/ValueSet/mimetypes%7C4.0.1)': include #0 has system urn:ietf:bcp:13 which could not be found, and the server returned error [Error from http://tx.fhir.org/r4: The code System "urn:ietf:bcp:13" has a grammar, and cannot be enumerated directly]; None of the provided codes ['#application/fhir+json'] are in the value set '[http://hl7.org/fhir/ValueSet/mimetypes|4.0.1](http://hl7.org/fhir/ValueSet/mimetypes%7C4.0.1)')

**Test Case 7.3.11**
[...]

Issue #523, opened 2024-06-21
Error: mimetype not found


SavannahDearing:
Error reported:
DiagnosticReport/fc5054c7-19d1-4d62-acd3-8862230c6548: DiagnosticReport.presentedForm[0].contentType: The System URI could not be determined for the code '**application/pdf**' in the ValueSet '[http://hl7.org/fhir/ValueSet/mimetypes|4.0.1](http://hl7.org/fhir/ValueSet/mimetypes%7C4.0.1)': include #0 has system urn:ietf:bcp:13 which could not be found, and the server returned error [Error from http://tx.fhir.org/r4: The code System "urn:ietf:bcp:13" has a grammar, and cannot be enumerated directly]

I'm trying to figure out if this is a problem with the validator not able to retrieve the mimetypes, or if 'application/pdf' is no longer accepted.

Thanks!

Jammjammjamm:
Are you encountering this on inferno.healthit.gov?

SavannahDearing:
yes
https://inferno.healthit.gov/suites/g10_certification/35jmkqBeT1y#g10_certification-multi_patient_api
sorry, I should have added the link in my previous post!

Jammjammjamm:
That does indicate an issue with the validator rather than your system, but I'm not seeing that issue when I run tests right now. If you rerun your tests, do you still see the error?

SavannahDearing:
OK I retested and no longer see this message.  That was an easy "fix" ;D 

Thanks!~

Jammjammjamm:
Ok, we are still investigating this issue as we have seen it pop up periodically.

SavannahDearing:
FYI, catching it again today:
https://inferno.healthit.gov/suites/g10_certification/13CPfpXPb9U

PR #526, opened 2024-07-03
Add error filter for tx validator errors

dehall:
Quick fix for the tx.fhir.org validation errors that we have seen slip through, as in #523 and #524. Now any message containing `[Error from tx.fhir.org/r4:` will be ignored.

## Testing notes
Testing this is tricky since this error is flaky. I used https://mockoon.com/ to create a mock service that always returned a bad error, with the attached response. [validate.txt](https://github.com/user-attachments/files/16086942/validate.txt)  I got this response from validator.fhir.org using a DocumentReference with a bad mime type "junk/pdf". Note I replaced the sessionID field at the bottom with a mockoon template helper so the sessionID isn't consistent (the validator session startup job broke when every instance got back the same session ID)


arscan:
A bit of an aside but Mockoon's implementation for creating mock services is somewhat interesting to me as it something we sometimes need to do for client testing.  Might be worth looking at for ideas in that area.

Jammjammjamm (approved):

PR #533, opened 2024-07-18
FI-2907: Asymmetric auth

Jammjammjamm:
This branch adds another launch group for confidential asymmetric client auth. In addition to testing this new launch group, the behavior of the other additional launches should also be verified.


lib/onc_certification_g10_test_kit/onc_program_procedure.yml
```diff
@@ -745,28 +745,41 @@ procedure:
...
+          * “client_id” (to support “client-public” “SMART on FHIR®
+            Capability”);
+          * Authorization header including “client_id” and “client_secret” (to
+            support “client-confidential-symmetric” “SMART on FHIR®
+            Capability”); and
+          * Authentication JSON Web Token (to support
+            “client-confidential-asymmetric” “SMART on FHIR® Capability”)
         inferno_supported: 'yes'
         inferno_tests:
-          - 1.3.05
```
yunwwang, line 768: Should we remove reference to SMART v1 test since this test apply to both SMART v1 and v2.

Jammjammjamm: `AUT-PAT-30` is SMART v2 only.

yunwwang:
Since AUT-PAT-30 is SMART v2 only, corresponding test 9.12 should aslo be SMART v2 only test.

yunwwang (approved):

PR #539, opened 2024-08-06
FI-2909: Add scope selection tests

Jammjammjamm:
This branch adds tests to verify that systems allow users to select granular scopes when resource-level scopes are requested.

Use the preset but run against qa. The scope selection tests should pass if you unselect the requested resource-level scopes and select granular scopes for Condition and Observation.


lib/onc_certification_g10_test_kit/smart_granular_scope_selection_group.rb
```diff
@@ -0,0 +1,193 @@
...
+
+module ONCCertificationG10TestKit
+  class SmartGranularScopeSelectionGroup < Inferno::TestGroup
+    title 'SMART Granular Scope Selection'
+    short_title 'SMART Granular Scope Selection'
+    id :g10_smart_granular_scope_selection
+
+    description <<~DESCRIPTION
+      These tests verify that when resource-level scopes are requested for
+      Condition and Observation resources, the user is prevented with the option
```
arscan, line 12:
```suggestion
      Condition and Observation resources, the user is presented with the option
```

lib/onc_certification_g10_test_kit/smart_granular_scope_selection_group.rb
```diff
@@ -0,0 +1,193 @@
...
+      of approving sub-resource scopes rather than the resource-level scope.
+
+      The first group requests v1 resource-level Condition and Observation
+      scopes, and the second group requests v2 resource-level Condition and
+      Observation socpes. In each instance, the user must unselect the
```
arscan, line 17:
```suggestion
      Observation scopes. In each instance, the user must unselect the
```

yunwwang (approved):

PR #540, opened 2024-08-06
FI-2982: Reorganize tests

Jammjammjamm:
This branch moves the visual inspection tests into their own top-level group, and rewords the titles for the additional tests group and its subgroups.

If this looks good, the test procedure and mapping spreadsheet will need to be updated.


arscan (approved):
I think this is a solid improvement that shouldn't be too disruptive.  But definitely something we'll want to point out in any materials we use to describe the changes.

PR #541, opened 2024-08-06
FI-2981: Update matrix generation for deeply nested tests

Jammjammjamm:
This branch updates the matrix generation to handle tests at different levels of nesting.

If you run `bundle exec rake g10_test_kit:generate_matrix` and examine the newly generated matrix, you should see:

- On the "Matrix" sheet, `9.14` should be identified as satisfying `AUT-PAT-28`.
- On the "Inferno Tests" sheet, all of the tests under `9.14` should appear.

This branch is on top of #537 because that PR contains the deeply nested tests which this change supports.

### DIFF
--- a/lib/onc_certification_g10_test_kit/tasks/generate_matrix.rb
+++ b/lib/onc_certification_g10_test_kit/tasks/generate_matrix.rb
@@ -50,6 +50,10 @@ module ONCCertificationG10TestKit
         workbook.write(FILE_NAME)
       end
 
+      def all_descendant_tests(runnable)
+        runnable.tests + runnable.groups.flat_map { |group| all_descendant_tests(group) }
+      end
+
       def generate_matrix_worksheet # rubocop:disable Metrics/CyclomaticComplexity
         matrix_worksheet = workbook.worksheets[0]
         matrix_worksheet.sheet_name = 'Matrix'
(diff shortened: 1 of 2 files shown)


lib/onc_certification_g10_test_kit/tasks/generate_matrix.rb
```diff
@@ -204,83 +205,75 @@ def applicable_options(runnable)
...
-        inferno_worksheet = workbook.worksheets[2]
+      def inferno_worksheet
+        workbook.worksheets[2]
+      end
 
-        columns = [
+      def columns # rubocop:disable Metrics/CyclomaticComplexity
+        @columns ||= [
           ['', 3, ->(_test) { '' }],
           ['', 3, ->(_test) { '' }],
```
arscan, line 215:
```suggestion
          ['Inferno Group', 13, ->(_test) { '' }],
```

A minor thing because it seemed a tad strange to me to not have a column name in here. I'm not sure I love how we flatten all groups into this column, but i can't think of an obvious better solution at the moment and this certainly works.

Screenshot 2024-08-06 at 4 10 51 PM

arscan (approved):
Just a minor formatting suggestion, take it or leave it.

I generated it and took a very quick look and it seems ok.  I also verified the 2 things you asked to verify in the PR.

PR #544, opened 2024-08-09
FI-2940: Test procedure

Jammjammjamm:
When I regenerate the matrix and open the spreadsheet, I get an error from excel. I don't know if that is because the yaml file refers to some tests which aren't on this branch (you'll see the errors when you regenerate the matrix).


yunwwang:
Test 10.47.01 is not shown for US Core 6.1.0 in excel. Regenerate matrix fixes that.

yunwwang (approved):

Issue #553, opened 2024-08-28
Automated token introspection tests should not be required for SMART v1


Jammjammjamm:
This issue was split from https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/550#issuecomment-2307147981

In release 6.0.0, the attestation for token introspection support was removed, and automated tests for token introspection were added. This change should only have applied for SMART App Launch STU2. When using SMART App Launch STU1, the attestation should still be used instead of the automated tests since SMART App Launch STU1 does not require a standard introspection implementation.

Jammjammjamm:
We released version 6.0.1 yesterday which should address this issue.

PR #554, opened 2024-08-28
FI-3049: Remove automated token introspection tests for SMART v1

Jammjammjamm:
See #553. When using SMART App Launch v1, this branch removes the automated token introspection tests and restores the token introspection attestation.


lib/onc_certification_g10_test_kit/visual_inspection_and_attestations_group.rb
```diff
@@ -189,6 +189,44 @@ class VisualInspectionAndAttestationsGroup < Inferno::TestGroup
       end
     end
 
+    test do
+      required_suite_options G10Options::SMART_1_REQUIREMENT
+      title 'Health IT developer demonstrated the ability of the Health IT Module / ' \
+            'authorization server to validate token it has issued.'
+      description %(
+        Health IT developer demonstrated the ability of the Health IT Module /
+        authorization server to validate token it has issued
```
arscan, line 198: Could we just add here something like:  "This is a functional requirement that requires manual inspection because SMART App Launch STU1 does not specify a standard approach to token introspection."

Just so we have somewhere "in the code (and UI)" that describes why SMART v1 does not require token introspection as an additional hint why this works as it does.

arscan (approved):
I verified that this removes the test from SMART v1 and replaces with the attestation.

PR #555, opened 2024-08-28
Release 6.0.1

Jammjammjamm:


arscan (approved):
I think you need to regenerate the procedure matrix too, but otherwise this seems fine.

PR #570, opened 2024-09-30
FI-3100: Update (g)(10) Certfication Test Kit Terminology Package to 2024 Version 

emichaud998:
This PR updates the related terminology script code to use the new 2024 UMLS terminology package version. Running the script now will pull down the 2024 UMLS package, and update the (g)(10) Test Kit to use the updated FHIR terminology. This also updates the `lib/inferno/terminology/expected_manifest.yml` file with the new number of values in the updated value sets so that the `bundle exec rake terminology:check_built_terminology` runs without failure. 

Also created instructions for this process and added it as an attachment to the ticket for this task.


lib/inferno/terminology/fhir_package_manager.rb
```diff
@@ -44,10 +45,14 @@ class << self
         #
         # @param [String] package The FHIR Package
         def get_package(package, destination, desired_types = [])
-          package_url = package
-            .split('#')
-            .prepend(REGISTRY_SERVER_URL)
-            .join('/')
+          package_url = if package.end_with?('#7.0.0')
```
Jammjammjamm, line 48: This should be a strict check for the US Core 7 package.

lib/inferno/terminology/tasks/download_fhir_terminology.rb
```diff
@@ -19,8 +19,9 @@ def download_us_core
           FHIRPackageManager.get_package('hl7.fhir.us.core#4.0.0', PACKAGE_DIR, ['ValueSet', 'CodeSystem'])
           FHIRPackageManager.get_package('hl7.fhir.us.core#5.0.1', PACKAGE_DIR, ['ValueSet', 'CodeSystem'])
           FHIRPackageManager.get_package('hl7.fhir.us.core#6.1.0', PACKAGE_DIR, ['ValueSet', 'CodeSystem'])
+          FHIRPackageManager.get_package('hl7.fhir.us.core#7.0.0', PACKAGE_DIR, ['ValueSet', 'CodeSystem'])
           FHIRPackageManager.get_package(
-            'us.nlm.vsac#0.11.0', File.join(PACKAGE_DIR, 'vsac'), ['ValueSet', 'CodeSystem']
+            'us.nlm.vsac#0.17.0', File.join(PACKAGE_DIR, 'vsac'), ['ValueSet', 'CodeSystem']
```
Jammjammjamm, line 24: This is not the version of the VSAC package referenced in US Core 7.

Jammjammjamm: You'll also have to regenerate everything from scratch.

Jammjammjamm: Run the g10 tests and verify that things look normal compared with what's on prod.

emichaud998: Re-ran with the vsac package update and everything looks normal and works as expected

Jammjammjamm:
Now you have to rerun the terminology generation.

emichaud998:
> Now you have to rerun the terminology generation.

Running now, will let you know if there are any issues

emichaud998:
Ran the terminology generation again and everything looks good and works as expected

Jammjammjamm (approved):

PR #576, opened 2024-10-21
FI-3101: Integrate US Core v7 into g(10) Test Kit

emichaud998:
Add US Core version 7 to the g(10) test kit. This PR includes several updates various test groups to include US Core 7 versions of each test that require a US Core suite version, and it includes a new Single Patient API test for US Core 7. This PR also adds new presets for this version, and updates spec tests to work with new changes. Note: Although Location was added as a direct mapping in USCDI v4, it was not added to the restricted or unrestricted resource type access tests since it can not directly be queried via a patient reference.


lib/inferno/terminology/loader.rb
```diff
@@ -199,7 +199,8 @@ def save_new_value_set_list
             USCoreTestKit::USCoreV311::USCoreTestSuite.metadata +
             USCoreTestKit::USCoreV400::USCoreTestSuite.metadata +
             USCoreTestKit::USCoreV501::USCoreTestSuite.metadata +
-            USCoreTestKit::USCoreV610::USCoreTestSuite.metadata
+            USCoreTestKit::USCoreV610::USCoreTestSuite.metadata +
+            USCoreTestKit::USCoreV700::USCoreTestSuite.metadata
```
Jammjammjamm, line 203: This belongs in the terminology update PR. It's what updates the file in my last comment on that PR.

emichaud998: Okay I will remove this here. I am currently working on addressing that last comment on the terminology PR and once I finish that, I will fix this here. I also see that I need to fix some linting issues

yunwwang (approved):

PR #580, opened 2024-11-01
FI-3401: Fix terminology

Jammjammjamm:
Rerun the terminology build and the rake task should report success. The test should pass as usual.


emichaud998 (approved):
Looks good, ran the terminology loader again and made sure rake tasks passed and everything looked correct when running test kit

PR #591, opened 2024-11-19
Release 7.0.1

yunwwang:
Release 7.0.1 incoporate US Core v0.9.2


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 7.0.1
+* Fix a bug which casued Inferno testing crash when Reference.reference element is not populated (#589)
```
arscan, line 2:
```suggestion
* Fix a bug which caused a system error when the `Reference.reference` element is not populated in certain resources (#589)
```

Gemfile.lock
```diff
@@ -325,8 +325,8 @@ GEM
       concurrent-ruby (~> 1.0)
     unicode-display_width (2.5.0)
     unicode_utils (1.4.0)
-    us_core_test_kit (0.9.1)
-      inferno_core (>= 0.4.37)
+    us_core_test_kit (0.9.2)
+      inferno_core (>= 0.5.0)
```
arscan, line 329: I find this a little surprising on a patch release but if @Jammjammjamm is good with this (I assume it was required here), no objections.

yunwwang: Before releasing 0.9.0, inferno_core dependency was updated to 0.5.0 using `bundle update` command. But the `gemspec` was not updated in that release.  I think it would be good to keep `gemspec` updated also.

Jammjammjamm: Bumping it in the gemspec wasn't strictly necessary, but it was also one of the test kits that had a problem that was addressed in 0.5.0.

Jammjammjamm (approved):

PR #598, opened 2024-12-17
FI-3428: Prevent systems from passing with US core 7 and SMART App Launch 1

Jammjammjamm:
This branch adds a failing test to the discovery group in the standalone patient app group when US Core 7 and SMART App Launch 1 are selected.

Screenshot 2024-12-18 at 9 39 14 AM

### DIFF
--- a/lib/onc_certification_g10_test_kit/smart_standalone_patient_app_group.rb
+++ b/lib/onc_certification_g10_test_kit/smart_standalone_patient_app_group.rb
@@ -96,6 +96,12 @@ module ONCCertificationG10TestKit
 
         id :g10_us_core_7_smart_version_check
         title 'US Core 7 requires SMART App Launch STU2+'
+        description %(
+          The [US Core 7 SMART on FHIR Obligations and
+          Capabilities](https://hl7.org/fhir/us/core/scopes.html) require SMART
+          App Launch STU2+, so systems can not certify with US Core 7 and SMART
+          App Launch STU1.
+        )
 
         run do
           assert false, 'US Core 7 is not eligible for certification with SMART App Launch STU1.'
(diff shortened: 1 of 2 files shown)


arscan:
You could perhaps also add in the description:

The [Test Procedure](https://www.healthit.gov/test-method/standardized-api-patient-and-population-services) also states in **Paragraph (g)(10)(v)(A) – Authentication and authorization for patient and user scopes**:

> *Note: US Core 7.0.0 must be tested with SMART App Launch 2.0.0 or above.*

arscan:
Looks like you need to regenerate the short ids.  I checked the test procedure to see if there was any mapping there that made sense.  Its not a distinct step, just a note above a bunch of steps.  So i guess you could map it to AUT-PAT-1 - 21.  I'm not sure if its worth it though.

yunwwang (approved):

lib/onc_certification_g10_test_kit/smart_standalone_patient_app_group.rb
```diff
@@ -90,6 +90,32 @@ class SmartStandalonePatientAppGroup < Inferno::TestGroup
...
            }
+
+      test do
+        required_suite_options(G10Options::US_CORE_7_REQUIREMENT)
+
+        id :g10_us_core_7_smart_version_check
+        title 'US Core 7 requires SMART App Launch 2.0.0 or above'
+        description %(
+          The [US Core 7 SMART on FHIR Obligations and
+          Capabilities](https://hl7.org/fhir/us/core/scopes.html) require SMART
```
yunwwang, line 101: Should we use version sepcific url here https://hl7.org/fhir/us/core/STU7/scopes.html

PR #605, opened 2025-01-07
[DOCS] Fix documentation footer and sidebar

arscan:
Unfortunately the footer and sidebar were committed to the wrong directory.  Perhaps we should have a convention for docs-only PRs so they update the wiki proactively (e.g.; by prepending the PR with [DOCS]), so we can see the change before merging so we don't end up in this situation again.  Its a tad risky, but I'm finding that not being able to see the change in the wiki it is hard to get right on the first try


Jammjammjamm:
> Perhaps we should have a convention for docs-only PRs so they update the wiki proactively (e.g.; by prepending the PR with [DOCS])

I think that is a reasonable solution.

Jammjammjamm (approved):

PR #631, opened 2025-03-11
Enhance docker services

daniel-alistar:
Sometimes, `inferno` is not ready when `nginx` starts and crashes because of that.
This fixes this unexpected behaviour.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,6 +21,8 @@ services:
       service: nginx
     volumes:
       - ./config/nginx.conf:/etc/nginx/nginx.conf
+    depends_on:
+      - inferno
   redis:
     extends:
       file: docker-compose.background.yml


arscan:
Hi @daniel-alistar --thanks for the suggestion.  While this establishes a dependency between the `nginx` container and the main `inferno` web service container, it won't cause the `nginx` container to wait for the `inferno` container to start accepting connections before giving up.  I do think it makes sense to establish that link (if `nginx` crashes if `inferno` isn't up, clearly a dependency exists), but I'm not sure that this will solve the specific problem you are seeing generically.  I think some kind of health check would have to be established.

I could be wrong though.  Has this helped in practice, from what you have seen?

docker-compose.yml
```diff
@@ -21,6 +21,8 @@ services:
       service: nginx
     volumes:
       - ./config/nginx.conf:/etc/nginx/nginx.conf
+    depends_on:
+      - inferno
```
kojiromike, line 25: You can solve this generically, by adding a [healthcheck](https://docs.docker.com/reference/compose-file/services/#healthcheck) to the inferno service, and then changing this `depends_on` to…

```suggestion
    depends_on:
      inferno: service_healthy
```

BlindChickens:
I have the same problem and is solved with the depends-on. 100% reproducable.

docker-compose.yml
daniel-alistar: This is not supported in docker compose version 3. ([documentation](https://docs.docker.com/compose/releases/release-notes/#compose-file-version-21-and-up-4))

At most, what we can do is to use `service_started` which guarantees that at least the container started before starting up `nginx`.

```suggestion
    depends_on:
      inferno:
        condition: service_started
```

PR #634, opened 2025-03-28
FI-3913 Fix indent calculator in matrix generator that broken when wrapped list items included

arscan:
I noticed this issue on attestations at the very end, but it occurs throughout.  Whenever there is a wrapped list item in markdown, the indentation-calculator messed up and improperly cut out leading characters in lines in the g10 matrix on the 'Inferno Tests' tab.

For example:

```
   A list is below
   * blah this is a list
     that spans multiple lines
   * it would break because it
     calculated indents on these second lines
```

Before:
Screenshot 2025-03-28 at 2 27 27 PM

After:
Screenshot 2025-03-28 at 2 25 04 PM

I updated the regenerated matrix as well.

I originally thought it had to do with the block line method differences on new attestations, which is why i changed that originally, but it wasn't that.  I left it to be consistent with all the others.


arscan:
Oh geez, this 'fix' is fixing it by mistake... it wasn't quite the same problem that i thought it was. I'm going to close this and put in a cleaner fix.

PR #664, opened 2025-06-03
build(terminology): use compose profiles to simplify terminology builder

kojiromike:
Use [docker compose profiles](https://docs.docker.com/compose/how-tos/profiles/) to avoid needing users to specify a distinct docker compose file for building terminology. Also, update the README to avoid an orphan container at the end of the terminology build process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,8 @@ Once that file exists, you can run the terminology creation task by using the
 following command:
 
 ```shell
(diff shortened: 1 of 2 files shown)


kojiromike:
(Also note that this is fully backwards compatible – it will work as well as ever for someone who uses the `terminology_compose.yml` directly.)

PR #761, opened 2026-07-09
Update US Core Gem version

ljtucker:
Update us core version to ensure the version makes it to the July 2026 Inferno release


karlnaden (changes requested):
bump the g10 version as well and change log. No need to have two PRs for this change.

karlnaden (approved):

